PR #11, opened 2020-12-01
Idempotent configuration binding and support for health checks

autodidaddict:
0.15.x requires that all capability providers:
* Gracefully ignore attempts at binding actors that are already bound
* Support returning a health status in response to a health check

### DIFF
--- a/http-server/src/lib.rs
+++ b/http-server/src/lib.rs
@@ -8,7 +8,7 @@ extern crate log;
 
 extern crate actix_rt;
 
-use crate::generated::core::CapabilityConfiguration;
+use crate::generated::core::{CapabilityConfiguration, HealthResponse};
 use actix_web::dev::Body;
 use actix_web::dev::Server;
 use actix_web::http::{HeaderName, HeaderValue, StatusCode};
@@ -23,12 +23,12 @@ use std::collections::HashMap;
 use std::error::Error;
 use std::sync::Arc;
 use std::sync::RwLock;
-use wascc_codec::core::{OP_BIND_ACTOR, OP_REMOVE_ACTOR};
+use wascc_codec::core::{OP_BIND_ACTOR, OP_REMOVE_ACTOR, OP_HEALTH_REQUEST};
 use wascc_codec::{deserialize, serialize};
 
 const CAPABILITY_ID: &str = "wascc:http_server";
 const VERSION: &str = env!("CARGO_PKG_VERSION");
-const REVISION: u32 = 2; // Increment for each crates publish
+const REVISION: u32 = 3; // Increment for each crates publish
 
 const OP_HANDLE_REQUEST: &str = "HandleRequest";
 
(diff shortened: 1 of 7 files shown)


http-server/src/lib.rs
```diff
@@ -23,12 +23,12 @@ use std::collections::HashMap;
...
 use std::sync::Arc;
 use std::sync::RwLock;
-use wascc_codec::core::{OP_BIND_ACTOR, OP_REMOVE_ACTOR};
+use wascc_codec::core::{OP_BIND_ACTOR, OP_REMOVE_ACTOR, OP_HEALTH_REQUEST};
 use wascc_codec::{deserialize, serialize};
 
 const CAPABILITY_ID: &str = "wascc:http_server";
 const VERSION: &str = env!("CARGO_PKG_VERSION");
-const REVISION: u32 = 2; // Increment for each crates publish
+const REVISION: u32 = 3; // Increment for each crates publish
```
brooksmtownsend, line 31: Doing some thinking on version v.s. revision, what purpose does revision serve if we properly use semver? I'd imagine each crates publish would correspond with a new semver, even if it's just a patch version.

brooksmtownsend (approved):
All looks good to me, looking for discussion on revision vs version

http-server/src/lib.rs
autodidaddict: There's a couple things going on here:

1. We shouldn't be maintaining revision inside the code anymore since providers now have claims. In fact, we're no longer even invoking the get descriptor function
1. Revision is used to provide a simple comparison operator "is this file newer than this other file?", this gives us the ability to not have to deal with parsing semver or dealing with labels that end in `-please-cthulhu-work-this-time`. 

tl;dr when we support hot-swapping, we'll require that the revision of the new file is greater than the revision of the old file, etc.

autodidaddict: I'm pretty sure both Android and iOS SDKs use this pattern, where they have a human-readable semver string and a monotonically increasing revision number for discrete comparison.

brooksmtownsend: Okay that makes sense, I see the value in having both now. I do remember that from the iOS days, you had to be very careful to remember to bump your revision or your entire build was moot (since it wouldn't be accepted).

I would like to see with hot-swapping the ability to go back in revision, so that we can support rollbacks. I suppose you could sign the old version with a new revision, curious on your thoughts there.

autodidaddict: we could just make the rule that you can't hot swap revision `n` with revision `n`, and all other swaps (up or down) are allowed. The main reason behind blocking in-place hot swaps is as a security measure to prevent tampered files from being injected.

autodidaddict:
Added idempotency and health check upgrades for redis and NATS. Going to create an issue for the other providers to all be upgraded to generated code, health checks, idempotency, and the stop function.

PR #21, opened 2021-01-20
Redisgraph use actor interfaces

brooksmtownsend:
This PR uses the graphdb interface from https://github.com/wasmCloud/actor-interfaces/pull/20/files as well as the `actor-core` interface to rewrite the redisgraph provider and actor API.


brooksmtownsend:
Pending merge https://github.com/wasmCloud/actor-interfaces/pull/20 and crates.io releases, I do have `git` dependencies scattered around here.

redisgraph/crates/wasccgraph-actor/Cargo.toml
```diff
@@ -1,18 +1,20 @@
 [package]
-name = "wasccgraph-actor"
-version = "0.0.3"
+name = "redisgraph-actor"
```
autodidaddict, line 2: I'm confused here. If we have the actor interface in `actor-interfaces/graphdb` , then why do we have another graph actor API that's specific to redis?

We should have a vendor-neutral graphdb actor interface in the actor-interfaces repository, and then we should have the stuff that _only_ applies to Redis in the redisgraph repository (the capability provider implementation).

redisgraph/crates/wasccgraph-actor/src/conversions.rs
```diff
@@ -0,0 +1,122 @@
+use crate::client_type_error;
```
autodidaddict, line 1: Again, this stuff belongs in the guest feature of the `actor_graphdb` crate, as none of this is Redis-specific...this abstraction that allows for the tuple-based query of a resultset should be available to all actors using graphdb, regardless of whether the implementation is Neo4j or redisgraph, etc.

autodidaddict (changes requested):
I think the only thing that seems out of place here is that the query wrapper (redisgraph-actor) code should be inside the actor interface, and not inside a redis-specific repository.

brooksmtownsend:
@autodidaddict ready for review here, tests failing because of relative dependencies, pending https://github.com/wasmCloud/actor-interfaces/pull/20 merge

autodidaddict (approved):

Issue #12, opened 2020-12-01
Upgrade all providers to 0.15.0 compliance


autodidaddict:
The redis, NATS, and http server providers are being upgraded since they're used with the integration tests. all other capability providers need to be upgraded to 0.15.x compatibility:

* Support health checks
* Ensure that binding operation is idempotent
* Support generated data types - do **NOT** use the wascc codec for anything other than the plugin trait.
* Support the `stop` function in the 0.9.0 codec / plugin definition and ensure that this stop function can't panic

brooksmtownsend:
Once #32 and #33 are closed, this issue can be closed.

PR #24, opened 2021-02-02
FS provider 0.15

brooksmtownsend:
Updated FS provider to use `wasmCloud` terminology and the `blobstore` actor interface for its shared types.

Tests are passing alongside the `actor-interface` doctest. The badges in the README will likely be corrupted for the time being until we get this provider published onto crates.io.

Related to #12

### DIFF
--- a/fs/src/lib.rs
+++ b/fs/src/lib.rs
@@ -24,17 +24,19 @@ mod chunks;
 #[cfg(not(feature = "static_plugin"))]
 capability_provider!(FileSystemProvider, FileSystemProvider::new);
 
+#[allow(unused)]
 const CAPABILITY_ID: &str = "wasmcloud:blobstore";
 const SYSTEM_ACTOR: &str = "system";
 const FIRST_SEQ_NBR: u64 = 0;
-const VERSION: &str = env!("CARGO_PKG_VERSION");
-const REVISION: u32 = 4; // Increment for each crates publish
+
+/// Tuple of (expected sequence number, chunks)
+type SequencedChunk = (u64, Vec<FileChunk>);
 
 #[derive(Clone)]
 pub struct FileSystemProvider {
     dispatcher: Arc<RwLock<Box<dyn Dispatcher>>>,
     rootdir: Arc<RwLock<PathBuf>>,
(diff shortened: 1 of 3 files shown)


fs/src/lib.rs
```diff
@@ -20,23 +18,23 @@ use std::{
...
 #[cfg(not(feature = "static_plugin"))]
 capability_provider!(FileSystemProvider, FileSystemProvider::new);
 
-const CAPABILITY_ID: &str = "wascc:blobstore";
+const CAPABILITY_ID: &str = "wasmcloud:blobstore";
 const SYSTEM_ACTOR: &str = "system";
 const FIRST_SEQ_NBR: u64 = 0;
 const VERSION: &str = env!("CARGO_PKG_VERSION");
-const REVISION: u32 = 3; // Increment for each crates publish
+const REVISION: u32 = 4; // Increment for each crates publish
```
autodidaddict, line 31: Revision and version are no longer needed for embedding in capability providers since we don't do descriptors.

fs/src/lib.rs
```diff
@@ -20,23 +18,23 @@ use std::{
...
 
+#[derive(Clone)]
 pub struct FileSystemProvider {
     dispatcher: Arc<RwLock<Box<dyn Dispatcher>>>,
-    rootdir: RwLock<PathBuf>,
-    upload_chunks: RwLock<HashMap<String, (u64, Vec<FileChunk>)>>,
+    rootdir: Arc<RwLock<PathBuf>>,
```
autodidaddict, line 36: If we don't actually need to use an `Arc` here, we should probably strip this. The litmus test I use is if we never need to clone this lock (which is required to pass it across a thread boundary), then we don't need to use an Arc.

autodidaddict (changes requested):

fs/src/lib.rs
brooksmtownsend: The reason for the `Arc` is so the FileSystemProvider can easily `derive(Clone)`, `RwLock` doesn't allow for cloning

autodidaddict: Yeah I forgot that a while back I made it so that the capability provider trait implementers had to be clone-able. Move along, nothing to see here. Please disperse!

autodidaddict (approved):

PR #30, opened 2021-02-11
Upgrading nats capability provider to 0.15 standards

autodidaddict:


nats/src/lib.rs
```diff
@@ -1,28 +1,28 @@
...
+extern crate wasmcloud_actor_core as actorcore;
+extern crate wasmcloud_actor_messaging as messaging;
+
 mod natsprov;
 
-const VERSION: &str = env!("CARGO_PKG_VERSION");
-const REVISION: u32 = 2; // Increment for each crates publish
+pub const VERSION: &str = env!("CARGO_PKG_VERSION");
+pub const REVISION: u32 = 3; // Increment for each crates publish
+pub const CAPABILITY_ID: &str = "wasmcloud:messaging"; // used by the Makefile
```
brooksmtownsend, line 11: `VERSION` and `REVISION` can be removed here, and `CAPABILITY_ID` can be marked `#[allow(unused)]` to address clippy warning

brooksmtownsend (changes requested):
One small request, otherwise LGTM.

nats/Makefile
```diff
@@ -8,7 +8,7 @@ NAME          := $(shell cargo metadata --no-deps --format-version 1 | jq -r '.p
 VERSION       := $(shell cargo metadata --no-deps --format-version 1 | jq -r '.packages[].version')
 LIBNAME       := $(shell echo $(NAME) | tr '-' '_')
 VENDOR        ?= wasmcloud
-REVISION	  ?= 0
+REVISION	  ?= 2
```
brooksmtownsend, line 11: This variable, if you check out the GH action, will be conditionally updated to the # of the run that is taking place. The `?=` will initialize the variable if it's not already defined in the environment already so it is needed. Basically, no harm in updating this, but it will be overwritten in the GH action so you might see a different value when this is published to wasmcloud.azurecr.io

brooksmtownsend (approved):

nats/Makefile
autodidaddict: Ah!! TIL something new. I can set it back to zero, or, as you say, it doesn't really matter

PR #38, opened 2021-03-03
Bumping NATS kvcache to provider core

autodidaddict:
When fixing clippy warnings in wasmcloud-host PR, decided to bump wasmcloud-host to use provider core instead of wascc-codec. This meant that I needed the NATS kvcache to use provider core ;)


brooksmtownsend (approved):
shipit

PR #56, opened 2021-03-18
Blobstore provider update

brooksmtownsend:
Both blobstore providers (fs, s3) needed to update to the blobstore interface version `0.2.0`, mostly just to return the `BlobstoreResult` for a few operations.

While I was updating and testing the s3 provider, I added in an option to use an HTTP proxy, which is common in enterprise AWS environments. This should not affect any existing configurations as it's strictly optional.

### DIFF
--- a/fs/src/lib.rs
+++ b/fs/src/lib.rs
@@ -86,7 +86,10 @@ impl FileSystemProvider {
         let container = sanitize_container(&container);
         let cdir = self.container_to_path(&container);
         std::fs::remove_dir(cdir)?;
-        Ok(vec![])
+        serialize(BlobstoreResult {
+            success: true,
+            error: None,
+        })
     }
 
     fn start_upload(
@@ -100,10 +103,17 @@ impl FileSystemProvider {
             container: blob.container,
         };
         let blob = sanitize_blob(&blob);
-        info!("Starting upload: {}/{}", blob.container.id, blob.id);
         let bfile = self.blob_to_path(&blob);
-        std::fs::write(bfile, &[])?;
-        Ok(vec![])
+        serialize(match std::fs::write(bfile, &[]) {
+            Ok(_) => BlobstoreResult {
+                success: true,
+                error: None,
+            },
+            Err(e) => BlobstoreResult {
+                success: false,
+                error: Some(e.to_string()),
+            },
+        })
     }
 
     fn remove_object(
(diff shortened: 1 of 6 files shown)


fs/src/lib.rs
```diff
@@ -113,8 +123,16 @@ impl FileSystemProvider {
     ) -> Result<Vec<u8>, Box<dyn Error + Sync + Send>> {
         let blob = sanitize_blob(&blob);
         let bfile = self.blob_to_path(&blob);
-        std::fs::remove_file(&bfile)?;
-        Ok(vec![])
+        serialize(match std::fs::remove_file(&bfile) {
```
autodidaddict, line 126: I've just been doing a refactor to code like this that might make this look "more idiomatic".

You can replace a match that returns 2 values of the same type with a `.map_or_else` function, where the first closure is what you do with the error value and the second closure is what you do with the Ok value.

From one of my actors undergoing this refactor:
```rust
kv::add_room(id, description).map_or_else(
        |e| RoomAck::fail(&format!("{}", e)),
        |_v| RoomAck::success(),
    )
```

s3/README.md
```diff
@@ -12,3 +12,14 @@ If you want to statically compile (embed) this plugin into a custom host, then e
...
+## Configuration Values
+| Value | Description |
+| ----------- | ----------- |
+| REGION | AWS region to use (default `us-east-1`) |
+| ENDPOINT | AWS endpoint to use (default `s3.us-east-1.amazonaws.com` |
+| AWS_ACCESS_KEY | AWS access key for authentication |
+| AWS_SECRET_ACCESS_KEY | AWS secret access key for authentication |
+| AWS_TOKEN | AWS token for authentication (can be omitted if not needed for auth) |
+| TOKEN_VALID_FOR | AWS token lifetime (in seconds)|
+| HTTP_PROXY | Proxy URL to use with the S3 client |
```
autodidaddict, line 25: :+1:

s3/src/lib.rs
```diff
@@ -107,12 +107,21 @@ impl S3Provider {
...
     ) -> Result<Vec<u8>, Box<dyn Error + Sync + Send>> {
         let rt = tokio::runtime::Runtime::new().unwrap();
-        rt.block_on(s3::remove_bucket(
-            &self.clients.read().unwrap()[actor],
-            &container.id,
-        ))?;
-
-        Ok(vec![])
+        serialize(
+            match rt.block_on(s3::remove_bucket(
```
autodidaddict, line 111: see previous comment on making this simpler code by replacing the match with a `map_or_else`

s3/src/lib.rs
```diff
@@ -166,13 +178,22 @@ impl S3Provider {
...
         let rt = tokio::runtime::Runtime::new().unwrap();
-        rt.block_on(s3::remove_object(
-            &self.clients.read().unwrap()[actor],
-            &blob.container.id,
-            &blob.id,
-        ))?;
-
-        Ok(vec![])
+        serialize(
+            match rt.block_on(s3::remove_object(
```
autodidaddict, line 182: Ditto.

autodidaddict (changes requested):
Just one recommendation to refactor to simplify some expressions

fs/src/lib.rs
brooksmtownsend: Good call, I'll look at refactoring this 😃

brooksmtownsend:
@autodidaddict addressed comments 👍🏻

autodidaddict (approved):

PR #64, opened 2021-03-31
Brought telnet provider up to 0.15.0 standards

### LINKED ISSUE #12: Upgrade all providers to 0.15.0 compliance

### LINKED ISSUE #32: Upgrade telnet provider to 0.15.0 compliance
The `telnet` provider needs to:

- [ ] Support health checks
- [ ] Ensure that binding operation is idempotent
- [ ] Support generated data types in [actor-interfaces](https://github.com/wasmCloud/actor-interfaces/) - do NOT use the wascc codec for anything other than the plugin trait.
- [ ] Support the stop function in the 0.9.0 codec / plugin definition and ensure that this stop function can't panic
- [ ] Use the `wasmcloud:telnet` capability contract ID
- [ ] Update `Cargo.toml` and `README.md` to the standard that the other capabilities set, with README badges, license information, descriptions, etc.

brooksmtownsend:
Fixes #32, Fixes #12 

This PR brings the `telnet` provider up to `0.15.0` standards, including making use of the `telnet` actor interface, idempotent `bind` functionality, and properly removing listeners when actors are Removed.

Speaking to the last point, I wasn't able to get the `RemoveActor` op to be delivered to this provider. Perhaps @autodidaddict can shed some light on if this should happen when I run `wash ctl stop actor`, `wash ctl stop provider`, or if it only is delivered in another case.

### DIFF
--- a/telnet/src/lib.rs
+++ b/telnet/src/lib.rs
@@ -1,32 +1,26 @@
-#[macro_use]
-extern crate wasmcloud_provider_core as codec;
-
-use log::{info, trace};
-mod server;
-mod session;
-
-use codec::{
-    capabilities::{CapabilityProvider, Dispatcher, NullDispatcher},
-    core::{OP_BIND_ACTOR, OP_HEALTH_REQUEST, OP_REMOVE_ACTOR, SYSTEM_ACTOR},
-    deserialize, serialize,
-};
-use crossbeam::channel::Sender;
-use serde::{Deserialize, Serialize};
+use crossbeam_channel::Sender;
+use log::{debug, info, warn};
 use std::{
     collections::HashMap,
     error::Error,
+    net::TcpListener,
     sync::{Arc, RwLock},
 };
 use wasmcloud_actor_core::{CapabilityConfiguration, HealthCheckResponse};
-
-type MessageHandlerResult = Result<Vec<u8>, Box<dyn Error + Send + Sync + 'static>>;
+use wasmcloud_actor_telnet::{SendTextArgs, OP_SEND_TEXT};
+use wasmcloud_provider_core::{
+    capabilities::{CapabilityProvider, Dispatcher, NullDispatcher},
+    capability_provider,
+    core::{OP_BIND_ACTOR, OP_HEALTH_REQUEST, OP_REMOVE_ACTOR, SYSTEM_ACTOR},
+    deserialize, serialize,
+};
+mod server;
+mod session;
 
 #[allow(dead_code)]
 const CAPABILITY_ID: &str = "wasmcloud:telnet";
 
-pub(crate) const OP_SEND_TEXT: &str = "SendText";
-pub(crate) const OP_SESSION_STARTED: &str = "SessionStarted";
-pub(crate) const OP_RECEIVE_TEXT: &str = "ReceiveText";
+type MessageHandlerResult = Result<Vec<u8>, Box<dyn Error + Send + Sync + 'static>>;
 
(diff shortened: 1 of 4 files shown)


telnet/src/lib.rs
```diff
@@ -54,33 +50,60 @@ impl TelnetProvider {
...
-        let outbound = self.outbounds.read().unwrap()[&msg.session].clone();
-        outbound.send(msg.text).unwrap();
-        Ok(vec![])
+    fn send_text(
+        &self,
+        _actor: &str,
+        msg: SendTextArgs,
+    ) -> Result<Vec<u8>, Box<dyn Error + Sync + Send>> {
+        match self.outbounds.read().unwrap().get(&msg.session).clone() {
+            Some(outbound) => serialize(outbound.send(msg.text).is_ok()),
```
autodidaddict, line 104: Are we really serializing a boolean as the result here? If that's the contract in the widl then I think it needs to be changed because right now the message pack serialization requires a wrapper object, e.g. we can't return scalars from functions.

autodidaddict (changes requested):

telnet/src/lib.rs
brooksmtownsend: It does look like we are returning `bool`s here: https://github.com/wasmCloud/actor-interfaces/blob/main/telnet/telnet.widl#L4-L6. I wonder if this was an artifact of an older widl design.

FWIW, the send text operation does work, but I haven't tried to actually unwrap the result and use it. Should returning a scalar with msgpack cause a panic right away or when you try and access the value? A result type would be more useful in general in order to pass error messages / failures

autodidaddict: If we don't use the result of that operation, then that's fine - it won't matter what we use because we never attempt to deserialize it.

However, if we ever attempt to deserialize a "floating bool" that has no wrapping container, that bool won't be deserializable with messagepack as coded in the actor interface libraries. The same code that wraps the function types in a unary wrapper for input should be applied to output as well, automatically creating a wrapper "args"... but that's an enhancement for another day.

autodidaddict: The error will only occur if the other side of a conversation attempts to de-serialize the scalar via messagepack.

autodidaddict (approved):

telnet/src/lib.rs
brooksmtownsend: Before merging this, should I go ahead and address this before we go further? I saw the approval but I'm happy to at least change the result type in the WIDL and update it here.

autodidaddict: Yeah we can just update the widl to indicate a void. It won't change the `Vec<u8>` return type of the capability provider since all operations return that.

brooksmtownsend: As long as I remember correctly too, updating the return type to `void` just ends up `map`ping the return type to `()`, which is pretty neat since it won't affect here even if we do return a bool.

brooksmtownsend: Opened up https://github.com/wasmCloud/actor-interfaces/pull/75 to update the interface, will upgrade this provider to that interface version `0.2.0` once it's published. For now, going to go ahead and merge this.

PR #65, opened 2021-04-01
Blobstore interface parity

### LINKED ISSUE #63: Ensure all blobstore operations are in parity with interface
For the `fs` and `s3` provider, some operations are failing to pass the (de)serialization trip as the functions are expecting different structures than the interface is providing.

For example, the `wasmcloud:blobstore` `remove_object` operation (https://github.com/wasmCloud/actor-interfaces/blob/main/blobstore/rust/src/generated.rs#L80_L100) takes in the object ID and container ID, both strings, and serializes them into a `RemoveObjectArgs` struct, like so:
```
        let input_args = RemoveObjectArgs {
            id: id,
            container_id: container_id,
        };
```

This is the structure that is then serialized and used to call the `RemoveObject` operation on the host.

The FS provider's `remove_object` handler (https://github.com/wasmCloud/capability-providers/blob/main/fs/src/lib.rs#L119-L123), however, accepts:
```
OP_REMOVE_OBJECT => self.remove_object(actor, deserialize(msg)?),
.....
    fn remove_object(
        &self,
        _actor: &str,
        blob: Blob,
    ) -> Result<Vec<u8>, Box<dyn Error + Sync + Send>> {
```
where the `msg` is the `RemoveObjectArgs` struct above.

In sum, what we're expecting in the blobstore providers as _input_ is not matching up with what the blobstore interface specifies. I recently ensured that the blobstore providers (fs and s3) return the proper _output_ structure, but another pass needs to be made to make both of these providers compliant with the interface.

brooksmtownsend:
Fixes #63 

With our `widl` wrapper object changes, some of the parameters for the blobstore provider functions were not the same type as the interface specifies. This PR remediates that.

### DIFF
--- a/fs/Cargo.toml
+++ b/fs/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "wasmcloud-fs"
-version = "0.3.3"
+version = "0.4.0"
 authors = ["wasmCloud Team"]
 edition = "2018"
 homepage = "https://wasmcloud.dev"
--- a/fs/src/lib.rs
+++ b/fs/src/lib.rs
@@ -70,9 +70,9 @@ impl FileSystemProvider {
     fn create_container(
         &self,
         _actor: &str,
-        container: Container,
+        args: CreateContainerArgs,
     ) -> Result<Vec<u8>, Box<dyn Error + Sync + Send>> {
-        let container = sanitize_container(&container);
+        let container = Container::new(sanitize_id(&args.id));
         let cdir = self.container_to_path(&container);
         std::fs::create_dir_all(cdir)?;
         serialize(&container)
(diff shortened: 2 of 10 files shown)


autodidaddict (approved):

autodidaddict:
Looks like there's still a discrepancy and the test is failing:

```
---- test::test_upload stdout ----
thread 'test::test_upload' panicked at 'called `Result::unwrap()` on an `Err` value: "Failed to de-serialize: missing field `container_id`"', src/lib.rs:574:14
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace

---- test::test_upload_and_download stdout ----
thread 'test::test_upload_and_download' panicked at 'called `Result::unwrap()` on an `Err` value: "Failed to de-serialize: missing field `blob_id`"', src/lib.rs:509:14

```

autodidaddict (changes requested):
need to clear up the mismatch

autodidaddict (approved):

PR #74, opened 2021-05-19
Cors

matthewtgilbride:


stevelr:
@matthewtgilbride Thanks for the pull request. I think adding CORS support is a good idea. In addition to fixing the build error, I suggest a couple changes to the PR:
- change the config name from ALLOWED_ORIGINS to CORS_ALLOWED_ORIGINS
- add CORS_ALLOWED_METHODS and CORS_ALLOWED_HEADERS with similar comma-separated format

Shout out if you need help addressing the build error.

matthewtgilbride:
> @matthewtgilbride Thanks for the pull request. I think adding CORS support is a good idea. In addition to fixing the build error, I suggest a couple changes to the PR:
> 
> * change the config name from ALLOWED_ORIGINS to CORS_ALLOWED_ORIGINS
> * add CORS_ALLOWED_METHODS and CORS_ALLOWED_HEADERS with similar comma-separated format
> 
> Shout out if you need help addressing the build error.

Hey @stevelr , I'm happy to add allowed methods and headers as well.  Definitely need help figure out the build error.  I reached out in slack here: https://wasmcloud.slack.com/archives/C01KG6W0Q3A/p1621772090014600?thread_ts=1621426628.001400&cid=C01KG6W0Q3A

Ultimately my problem (I think) stems from the fact that I need to reference something from `cfgvals` inside of the `HttpServer::new` call, which requires `'static` lifetimes.  I'm fighting the borrow checker hard here....HALP! 😄

matthewtgilbride:
P.s., do we need to implement any logic around non-sensical input? (I guess `actix-cors` doesn't, so we need not either).

What I mean is....does it make any sense to provide `CORS_ALLOWED_ORIGINS` and _not_ provide any allowed methods?  I am no CORS expert but that feels like it would be the same as not allowing CORS at all (since no methods would be allowed).

matthewtgilbride:
Implemented methods and headers, ready for review 👍

autodidaddict (approved):

PR #78, opened 2021-06-03
bump httpserver for CORS Support

brooksmtownsend:
Thank you @matthewtgilbride for the PR #74 to introduce CORS support, this is just a PR that bumps the httpserver a patch version, as the configuration values are optional and don't change the current behavior.

### DIFF
--- a/http-server/Cargo.toml
+++ b/http-server/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "wasmcloud-httpserver"
-version = "0.12.2"
+version = "0.12.3"
 authors = ["wasmCloud Team"]
 edition = "2018"
 homepage = "https://wasmcloud.dev"


stevelr:
rolling back http server to the previous actix version broke the http server provider. This will have to get resolved when the http provider gets updated for otp.

stevelr:
Will fix with http server in 0.50 release

brooksmtownsend:
@stevelr I think I lost my context here, should we still merge this in or should we hold off on releasing this httpserver version?

autodidaddict:
Agreed that I'm confused on context @stevelr - I think the provider for which this issue was originally created is deprecated in favor of the forthcoming 0.5x-based provider. Is this now just a flag to ensure that we have CORS support in the new provider?

stevelr:
works in new httpserver provider for otp. Marking fixed.

PR #79, opened 2021-07-24
Updated actix_web for http-server

pxjohnny:
When trying to run httpserver 0.12.2 on an M1 mac it crashes due to the use of actix_rt 2.x with actix_web 3.x.

### DIFF
--- a/http-server/Cargo.toml
+++ b/http-server/Cargo.toml
@@ -26,7 +26,7 @@ rmp-serde = "0.15.4"
 serde_bytes = "0.11.5"
 serde = { version = "1.0.124", features = ["derive"] }
 actix-cors = "0.5.4"
-actix-web = "3.3.2"
+actix-web = "4.0.0-beta.8"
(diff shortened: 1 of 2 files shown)


autodidaddict (commented):
Is there no way to make this work without depending on betas?

pxjohnny:
I don't think you can avoid the beta versions for now.. also if you check v0.12.1 it used actix_web 4.0.0 beta 4 as well...

brooksmtownsend:
@pxjohnny Have you tested this provider out with some current samples and verified it's working? If so, what testing did you do?

pxjohnny:
I tested my changes with the Echo actor (0.2.1) and a custom actor I created on an M1 MacBook Pro. You can find my compiled provider at container.bdexpress.rocks:5443/httpserver:0.12.9. I only compiled it for armv7-unknown-linux-gnueabihf, aarch64-apple-darwin and x86_64-unknown-linux-gnu for my testings.

PR #82, opened 2021-08-12
Httpserver and KVRedis capability providers for otp

stevelr:


kvredis/src/main.rs
```diff
@@ -0,0 +1,307 @@
...
+        let val: i32 = self.exec(ctx, &mut cmd).await?;
+        Ok(val)
+    }
+
+    /// Returns true if the store contains the key
+    async fn contains<TS: ToString + ?Sized + Sync>(
+        &self,
+        ctx: &Context,
+        arg: &TS,
+    ) -> RpcResult<bool> {
```
brooksmtownsend, line 121: For this and the other commands, I'm a little confused as to the purpose of the generic TS here. Is there a downside to just using a reference to a string slice `&str` to keep the method signature less complicated?

kvredis/src/main.rs
```diff
@@ -0,0 +1,307 @@
...
+        cmd.query_async(con.deref_mut()).await.map_err(to_rpc_err)
+    }
+}
+
+/// Handle incoming rpc messages and dispatch them to applicable trait handler.
+#[async_trait]
+impl MessageDispatch for KvRedisProvider {
+    async fn dispatch(&self, ctx: &Context, message: Message<'_>) -> RpcResult<Message<'_>> {
+        let op = match message.method.split_once('.') {
+            Some((cls, op)) if cls == "KeyValue" => op,
```
brooksmtownsend, line 291: Forgive me if I missed this in the interface, does this mean all of our operations contain a `.` now? e.g.
- KeyValue.Set
- KeyValue.Add

brooksmtownsend: As long as the operation names are available as constants to actors and providers, no changes needed. Just looking for clarification

kvredis/tests/kv_test.rs
```diff
@@ -0,0 +1,389 @@
...
+    let prov = test_provider().await;
+
+    // health check
+    let hc = prov.health_check().await;
+    check!(hc.is_ok())?;
+    Ok(())
+}
+
+/// get and set
+async fn get_set() -> RpcResult<()> {
```
brooksmtownsend, line 63: Love seeing integration tests for the providers. Much needed 🚀

brooksmtownsend (changes requested):
Only suggestion that doesn't fit into the regular PR comments: I'd suggest moving the README into `pre-otp` and writing a new one for this repository that at least has a little snippet saying that it's under construction, where to find the old capability providers, etc.

There's a lot of changes here so forgive me if my comments are irrelevant because I'm missing context. Overall, this is awesome and it's great to see these really well-thought-out implementations of providers and the new interfaces in action

httpserver-rs/bin/main.rs
```diff
@@ -0,0 +1,86 @@
...
+
+/// Handle RPC messages. This provider doesn't have any.
+impl ProviderDispatch for HttpServerProvider {}
+#[async_trait]
+impl MessageDispatch for HttpServerProvider {
+    async fn dispatch(
+        &self,
+        _ctx: &Context,
+        message: Message<'_>,
+    ) -> Result<Message<'_>, RpcError> {
```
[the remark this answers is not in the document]
stevelr: I did remove the lifetime parameter from Context, which is the only one most actors see, and the only thing we talked about in that phone call.
Lifetime on Message allows pass-by-reference of message args. so less copying, and I hope in the future will allow us to use zero-copy from the msgpack buffer all the way to the message handler. The motivation for removing it from actor code was to simplify the appearance for novice rust programmers. I expect that capability provider developers are more likely to be more experienced developers and that the lifetime won't be a signfiicant burden.

kvredis/src/main.rs
stevelr: The purpose is to allow you to use &str, String, or &String - it gives the most flexibility for the user of the api. It's all code-generated, so it doesn't add complexity to the developer either.

The secondary benefit, besides convenience for the caller, is reducing an allocation. If we made it &str, and the caller had passed &String, the function has to allocate another String because it needs to be owned further down. On the other hand, if the caller has a string they can pass it and no allocation is required.

kvredis/src/main.rs
stevelr: the benefit is that no constants are required. OP constants don't need to be declared, and you avoid the bugs we had before of misspelling them and then wondering why the message wasn't received. The "client" sdk adds the method name in, and the dispatcher uses it.

build/makefiles/provider.mk
```diff
@@ -0,0 +1,207 @@
...
+	cargo clean
+	cross clean
+endif
+
+
+
+# auto rust dependencies
+# TODO
+#   target: $(rust_deps)
+#     $(cargo) build --target $(...)
```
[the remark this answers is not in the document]
stevelr: done (next commit)

build/makefiles/provider.mk
```diff
@@ -0,0 +1,207 @@
...
+	@echo "platform       : $(platform)"
+	@echo "machine_id     : $(machine_id)"
+	@echo "default-par    : $(par_target0)"
+	@echo "project_dir    : $(project_dir)"
+	@echo "subdirs        : $(subdirs)"
+	@echo "top_targets    : $(top_targets)"
+	@echo "NAME           : $(NAME)"
+	@echo "VENDOR         : $(VENDOR)"
+	@echo "VERSION        : $(VERSION)"
+	@echo "REVISION       : $(REVISION)"
```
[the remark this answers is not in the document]
stevelr: I'd prefer to leave it in - not finished debugging makefiles yet. Do you think the target should be renamed?

httpserver-rs/Cargo.toml
```diff
@@ -0,0 +1,47 @@
...
+toml = "0.5"
+warp = {version="0.3", features=["tls"]}
+
+[dependencies.wasmcloud-interface-httpserver]
+version = "0.1"
+#path = "../../interfaces/httpserver/rust"
+
+[dependencies.wasmbus-rpc]
+version = "0.3"
+#path = "../../weld/rpc-rs"
```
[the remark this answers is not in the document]
stevelr: I'll remove them in the future when things are more stable. There are so many inter-crate dependencies that I need to be able to quickly uncomment those for testing.

kvredis/provider_test_config.toml
```diff
@@ -0,0 +1,27 @@
...
+#redis_url = "0.0.0.0:6379"
+
+# lattice prefix (default "default")
+#lattice_rpc_prefix = "default"
+
+# link name (default: "deefault")
+#link_name = "default"
+
+# name of contract under test
+contract_id = "wasmcloud:keyvalue"
```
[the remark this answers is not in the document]
stevelr: wow, so many typos in there. 
There's a method to the madness. Commented-out lines should be set to the default value. Any setting not commented either overrides the default, or defines a setting that has no default.

build/makefiles/provider.mk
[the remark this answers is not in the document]
brooksmtownsend: I think it's fine for now

kvredis/src/main.rs
brooksmtownsend: Okay, the real thing that I think makes this fine is that it's all codegen 👍🏻 Thanks for the answer

brooksmtownsend (approved):
Changes look great, looks like there's just some merge conflicts to deal with

README.md
```diff
@@ -1,75 +1,17 @@
-# capability-providers
+This repository contains capability providers for wasmcloud.
```
autodidaddict, line 1: Please add a markdown header for this README, something like `# Capability Providers`

build/makefiles/provider.mk
```diff
@@ -0,0 +1,198 @@
...
+
+target/%/release/$(bin_name): $(RUST_DEPS)
+	tname=`echo -n $@ | sed -E 's_target/([^/]+)/release.*$$_\1_'` &&\
+	cross build --release --target $$tname
+	@rm -f $(dest_par)
+
+endif
+
+# make exs stub for provider
+build/stub.exs: $(dest_par)
```
autodidaddict, line 128: Can you explain what this is? No one should be required to have Elixir installed or run Elixir in order to build, test, or create their own capability provider. If this is just a handy thing for wasmcloud-host developers to have while testing, then we should indicate this somewhere, either in a comment or in "help" output from the makefile, or both. 

If I don't know why we're emitting a `stub.exs` file then I don't expect any other developer will, either.

build/makefiles/provider.mk
```diff
@@ -0,0 +1,198 @@
...
+		name: $(NAME),
+		path: "$(abspath $(dest_par))",
+		key: "$(shell wash par inspect $(dest_par) -o json | jq -r ".service")",
+		link: "default",
+		contract: "$(CAPABILITY_ID)",
+	},
+	EOF
+
+
+# is this needed for cap providers?
```
autodidaddict, line 141: If this is commented out because it's not needed, then we should remove these lines from the file.

httpserver-rs/Cargo.toml
```diff
@@ -0,0 +1,47 @@
...
+rmp-serde = "0.15"
+serde_bytes = "0.11"
+serde_json = "1.0"
+serde = {version = "1.0", features = ["derive"] }
+thiserror = "1.0"
+tokio = { version = "1", features = ["full"] }
+toml = "0.5"
+warp = {version="0.3", features=["tls"]}
+
+[dependencies.wasmcloud-interface-httpserver]
```
autodidaddict, line 29: convert this to a regular dependency declaration, e.g. `wasmcloud-interface-httpserver = "0.1"`

httpserver-rs/Cargo.toml
```diff
@@ -0,0 +1,47 @@
...
+version = "0.1"
+#path = "../../interfaces/httpserver/rust"
+
+[dependencies.wasmbus-rpc]
```
autodidaddict, line 33: convert to a single-line dependency declaration.

httpserver-rs/Makefile
```diff
@@ -0,0 +1,18 @@
...
+CAPABILITY_ID = wasmcloud:httpserver
+NAME = "HttpServer"
+VENDOR = "wasmCloud"
+PROJECT = httpserver
+# first two-parts of the semver version
+VERSION = 0.1
+# the tag in the registry will be VERSION.REVISION, and must be incremented
+# any time the binary changes.
+# During development iterations, REVISION should be incremented each build
+REVISION = 5
```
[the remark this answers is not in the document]
autodidaddict: Agree with Brooks here. This is not the versioning scheme we use for our publicly published capability providers in AzureCR.

httpserver-rs/bin/main.rs
```diff
@@ -0,0 +1,86 @@
...
+use tokio::sync::RwLock;
+use wasmbus_rpc::{core::LinkDefinition, provider::prelude::*};
+use wasmcloud_provider_httpserver::{load_settings, HttpServer};
+
+fn main() -> Result<(), Box<dyn std::error::Error>> {
+    // handle lattice control messages and forward rpc to the provider dispatch
+    // returns when provider receives a shutdown control message
+    provider_main(HttpServerProvider::default())?;
+
+    eprintln!("HttpServer provider exiting");
```
autodidaddict, line 14: Should this emit to `stderr` ? If this line executes as a result of a graceful exit, we should make it explicit that this isn't a "bad" exit.

kvredis/Cargo.toml
```diff
@@ -0,0 +1,46 @@
...
+redis = {version = "0.21", features = ["tokio-comp"] }
+rmp-serde = "0.15"
+serde_bytes = "0.11"
+serde_json = "1.0"
+serde = {version = "1.0", features = ["derive"] }
+thiserror = "1.0"
+tokio = { version = "1", features = ["full"] }
+toml = "0.5"
+
+[dependencies.wasmcloud-interface-keyvalue]
```
autodidaddict, line 25: convert to single-line and remove unnecessary comment

kvredis/Cargo.toml
```diff
@@ -0,0 +1,46 @@
...
+version = "0.2"
+#path = "../../interfaces/keyvalue/rust"
+
+[dependencies.wasmbus-rpc]
```
autodidaddict, line 29: Make this single-line

autodidaddict (changes requested):

build/makefiles/provider.mk
stevelr: This is only needed as a band-aid to help build elixir manifest files, which is not intended to be a supported feature, but something I used for development. When wash can process manifest files and send control-interface commands to the host, this will go away.

build/makefiles/provider.mk
stevelr: deleted. weld commands moving to wash.

httpserver-rs/bin/main.rs
stevelr: env_logger and pretty_env_logger both write to stderr by default - for all logs.- I was following that convention.

httpserver-rs/Makefile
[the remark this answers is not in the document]
stevelr: didn't intend to change the convention. I just misinterpreted it. Thanks for the clarification.

PR #84, opened 2021-08-19
Httpclient capability provider

stevelr:


httpclient/src/main.rs
```diff
@@ -0,0 +1,179 @@
...
+        };
+        HttpClientReceiver::dispatch(
+            self,
+            ctx,
+            &Message {
+                method: op,
+                arg: message.arg,
+            },
+        )
+        .await
```
brooksmtownsend, line 177: Is this part codegenned?

brooksmtownsend (commented):
Two questions

httpclient/provider_test_config.toml
```diff
@@ -0,0 +1,26 @@
...
+rust_log = "debug"
+
+# set RUST_BACKTRACE (default: 0)
+rust_backtrace = "1"
+
+# nats should be running. Uncomment to override the default url
+#nats_url = "0.0.0.0:4222"
+
+# redis should be running. Uncomment to override the default url
+#redis_url = "0.0.0.0:6379"
```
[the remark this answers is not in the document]
stevelr: good catch - redis is not needed.
nats is needed for the test because the test communicates with the provider over rpc.

httpclient/src/main.rs
stevelr: not yet :-). There are still a couple of open issues that need to be nailed down before this can be code-genned:
- can a capability provider implement more than one interface? (I think yes)
- should messages have a "Service." prefix?
The latter is something I was going to test today. I believe actors and the providers I've built have support for it. The only thing that doesn't is the builtin providers.

brooksmtownsend: Okay, I was just wondering here 🙂 implementing more than one service is definitely something to discuss

brooksmtownsend (approved):

PR #85, opened 2021-08-24
nats messaging capability provider for otp - initial checkin, with unit test

stevelr:

### DIFF
--- a/nats/src/main.rs
+++ b/nats/src/main.rs
@@ -1,0 +1,305 @@
+//! Nats implementation for wasmcloud:messaging.
+//!
+
+use log::error;
+use nats::asynk::{self, Message as NatsMessage};
+use serde::{Deserialize, Serialize};
+use std::{collections::HashMap, convert::Infallible, sync::Arc};
+use tokio::sync::RwLock;
+use wascap::prelude::KeyPair;
+use wasmbus_rpc::{core::LinkDefinition, provider::prelude::*, MessageDispatch};
+use wasmcloud_interface_messaging::{
+    MessageSubscriber, MessageSubscriberSender, Messaging, MessagingReceiver, PubMessage,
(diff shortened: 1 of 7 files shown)


stevelr:
There is a known failing build check - it isn't finding the messaging interface crate. When the messaging interface PR is merged and the interface crate published, the build check for this provider will succeed.

nats/src/main.rs
```diff
@@ -0,0 +1,302 @@
...
+    }
+
+    /// send message to subscriber
+    async fn dispatch_msg(&self, ld: &LinkDefinition, nats_msg: NatsMessage) {
+        let msg = SubMessage {
+            body: nats_msg.data,
+            reply_to: nats_msg.reply,
+            subject: nats_msg.subject,
+        };
+        let tx = wasmbus_rpc::provider::ProviderTransport {
```
autodidaddict, line 149: We don't need to do this for MVP, but similar to how the interface was simplified, is there any way to condense the use of the provider transport with the `MessageSubscriberSender`, even if the only thing we do is avoid the manual construction of the struct with a `new`

autodidaddict (approved):

nats/src/main.rs
```diff
@@ -146,13 +146,8 @@ impl NatsMessagingProvider {
...
             subject: nats_msg.subject,
         };
-        let tx = wasmbus_rpc::provider::ProviderTransport {
-            bridge: get_host_bridge(),
-            ld,
-        };
-        let ctx = Context::default();
-        let actor = MessageSubscriberSender::via(tx);
-        if let Err(e) = actor.handle_message(&ctx, &msg).await {
+        let actor = MessageSubscriberSender::for_actor(ld);
```
autodidaddict, line 149: :+1:

autodidaddict (approved):

PR #89, opened 2021-09-03
HTTPClient release action

brooksmtownsend:
First iteration of release actions including template and httpclient, just to get started. Looking for any feedback before I replicate this for the rest of the providers


autodidaddict:
The rust check test is failing on this. @brooksmtownsend is it supposed to be?

brooksmtownsend:
@autodidaddict No it's not supposed to be failing, I just didn't include the logic to run nats alongside the rust check step so that it can be used in the integration tests. Re-running now, and 🤞🏻 should work fine

autodidaddict (approved):

.github/workflows/TEMPLATE_provider.yml
```diff
@@ -1,41 +1,64 @@
...
-      - run: rustup component add clippy
-      - name: Run cargo clippy
-        run: cargo clippy --all-features
-        working-directory: ${{env.working-directory}}
+      - uses: wasmcloud/common-actions/install-wash@main
+      - uses: wasmcloud/common-actions/install-cross@main
+      - name: Build full provider archive
+        run: make par-full
+        working-directory: ${{ env.working-directory }}
+      # Push artifact to https://AZURECR/provider:VERSION
```
stevelr, line 56: can push artifact move to a common action?

pre-otp/.github/workflows/httpclient.yml
```diff
@@ -0,0 +1,39 @@
+name: HTTPCLIENT
```
stevelr, line 1: I worry about the pre-otp workflows becoming out of date since they won't be maintained. Should we delete pre-otp workflows?

build/makefiles/provider.mk
```diff
@@ -102,16 +103,17 @@ $(dest_par): $(bin_target0) Makefile Cargo.toml
...
 par-full: $(dest_par) $(bin_targets)
-	# add other defined targets
 	for target in $(par_targets); do \
-	    target_dest=target/$$target/release/$(bin_name);  \
-	    par_arch=`echo -n $$target | sed -E 's/([^-]+)-([^-]+)-([^-]+)(-gnu)?/\1-\3/'`; \
+	    target_dest=target/$${target}/release/$(bin_name);  \
+		if [ $$target = "x86_64-pc-windows-gnu" ]; then \
+			target_dest=$$target_dest.exe;  \
+		fi; \
+	    par_arch=`echo -n $$target | sed -E 's/([^-]+)-([^-]+)-([^-]+)(-gnu.*)?/\1-\3/' | sed 's/darwin/macos/'`; \
```
stevelr, line 113: instead of doing the sed here inside this one rule, is there a variable we could change so that target is already correct?

stevelr (changes requested):

brooksmtownsend:
@stevelr sorry to merge before your comments, I'll open a new PR once we resolve the conversations here (I'll cancel the release)

pre-otp/.github/workflows/httpclient.yml
brooksmtownsend: We certainly can, I only kept them so they'd be around with the old providers. I can just remove them

stevelr:
a necessary hazard of working asynchronously in multi time zones

build/makefiles/provider.mk
brooksmtownsend: Yeah, I might want your help with that as there's a good bit of complexity in the Makefile. Because everything was generally correct other than the target, I wasn't sure if you matched against `x86_64-darwin` or `aarch64-darwin` anywhere else.

.github/workflows/TEMPLATE_provider.yml
brooksmtownsend: Yes, the way that would have to work is by giving it an input for the path to the artifact (e.g. `httpclient/build/httpclient.par.gz`) and a registry URL. Let me see what I can get working

PR #90, opened 2021-09-03
Httpclient fixes

brooksmtownsend:
Addressing some comments from @stevelr from #89 when I closed it too quickly.

- Removed old actions from the pre-otp repo
- Refactored the artifact release into a common action, that can be found [here](https://github.com/wasmCloud/common-actions/blob/main/oci-artifact-release/action.yml)

TODO: correct the architecture in the Makefile from `x86_64-darwin` to `x86_64-macos` rather than just using a sed in a single target. I'm not sure what will break if I just change this, so let me know if there's any gotchas @stevelr


stevelr (approved):
hmm - making the common action didn't reduce the number of lines, but it's config instead of code, so still an improvement!

PR #92, opened 2021-09-04
Implement remaining actions

brooksmtownsend:
This PR is essentially copy/filling in the other github actions for httpserver, kvredis, nats. I've added NATS for all 3, and then included redis as well for kvredis.

I've also adjusted the versions to indicate breaking changes over our current providers to avoid any possible confusion there.


stevelr (approved):
LGTM!

PR #94, opened 2021-09-07
Correct provider archive release

brooksmtownsend:
This PR re-arranges the release actions for capability providers a bit in order to run rust checks, build the provider archive, and THEN release the archive to azurecr and attach it to the github release. This also fixes a bug where the provider archive wasn't persisted between jobs.

I've tested these steps with a premade provider archive and you can see the results here: https://github.com/wasmCloud/capability-providers/releases/tag/testrelease


stevelr (commented):
Do these run on any tag push, or just annotated/release tags?

stevelr (approved):

brooksmtownsend:
> Do these run on any tag push, or just annotated/release tags?

These run on tag push, but only if it matches the proper format of the provider tag. For example, the httpclient provider release action will only trigger when a tag is pushed to main in the form of `httpclient-v*`

PR #96, opened 2021-09-09
add timeouts for httpserver

stevelr:
The default timeout for all http responses is 2 seconds. It can be overridden on a per-link basis by adding a value for "timeout_ms" to the settings. The units are milliseconds.

If any actor takes longer than the timeout to respond, the httpserver responds to the http client with an http status 503. I created a test case to confirm that the timeout works as expected.

Bump the version to 0.1.4.

This depends on wasmbus-rpc 0.3.11 which has already been published to crates.io


httpserver-rs/Cargo.toml
```diff
@@ -1,6 +1,6 @@
 [package]
 name = "wasmcloud-provider-httpserver"
-version = "0.1.3"
+version = "0.1.4"
```
brooksmtownsend, line 3: In the automation PR, I changed the version to `0.14.0` in order to be a minor change over the previously released httpserver, 0.13.0. I think this can be changed to `0.14.1` to stick with that

brooksmtownsend (changes requested):
all the code looks great, just a request to bump the version up to what's in `main`

httpserver-rs/src/lib.rs
```diff
@@ -263,16 +273,38 @@ impl HttpServer {
...
-        Ok(resp)
+        match actor.handle_request(&ctx, &req).await {
+            Err(RpcError::Timeout(_)) => {
+                error!(
+                    "actor {} req path {} timed out: returning 503",
+                    &ld.actor_id, &req.path,
+                );
+                Ok(HttpResponse {
+                    status_code: 503,
+                    body: b"Actor took too long or was unavailable.".to_vec(),
```
autodidaddict, line 287: I know this seems like a super nitpick, but for security reasons, maybe we shouldn't disclose that there's an actor on the other side of this provider. Maybe just "No response within timeout" or something like that.

autodidaddict (changes requested):
Assuming the version numbers all check out per @brooksmtownsend comments, this LGTM. I think for security reasons we shouldn't disclose to consumers that an actor is on the other side of this HTTP endpoint in the timeout error message.

httpserver-rs/src/lib.rs
autodidaddict: Getting this error message would immediately tell a consumer they've discovered a wasmCloud installation.

stevelr: good point

stevelr:
Addressed PR feedback. Also fixed https://github.com/wasmCloud/capability-providers/issues/99 in this branch.

httpserver-rs/src/settings.rs
```diff
@@ -176,9 +176,13 @@ impl ServiceSettings {
 ///   config_file: load from file name. Interprets file as json, toml, yaml, based on file extension.
 ///   config_b64:  base64-encoded json string
 ///   config_json: raw json string
+/// Also accept "address" (a string represengint SocketAddr) and "PORT", a localhost port
```
brooksmtownsend, line 179: representing is a typo here

httpserver-rs/src/settings.rs
```diff
@@ -195,6 +199,25 @@ pub fn load_settings(values: &HashMap<String, String>) -> Result<ServiceSettings
         settings.merge(ServiceSettings::from_json(str.as_bytes())?);
     }
 
+    // accept address as value parameter
+    if let Some(addr) = values.get("address") {
```
brooksmtownsend, line 203: By convention previously we've had all of the values be uppercase, if we want to stick with that I'd suggest changing this to `ADDRESS`.

I'm not a fan of changing everything in past providers, but I'd be OK with making values case insensitive. It would be up to the provider developer to do this though.

brooksmtownsend (changes requested):

brooksmtownsend (approved):
Love the sass, LGTM

PR #101, opened 2021-09-13
Derive macro provider dispatch

stevelr:
Use derive macro in providers.
new versions:
httpclient: 0.3.1
httpserver: 0.14.2
kvredis: 0.14.1
nats: 0.11.1

### DIFF
--- a/httpclient/tests/httpclient_test.rs
+++ b/httpclient/tests/httpclient_test.rs
@@ -43,7 +43,7 @@ async fn get_request(_opt: &TestOptions) -> RpcResult<()> {
     let ctx = Context::default();
 
     let resp = client
-        .request(&ctx, &HttpRequest::get("https://wttr.in/London?format=3"))
+        .request(&ctx, &HttpRequest::get("https://wttr.in/London"))
         .await?;
     assert_eq!(resp.status_code, 200, "status code");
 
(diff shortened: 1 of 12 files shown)


httpclient/tests/httpclient_test.rs
```diff
@@ -43,7 +43,7 @@ async fn get_request(_opt: &TestOptions) -> RpcResult<()> {
     let ctx = Context::default();
 
     let resp = client
-        .request(&ctx, &HttpRequest::get("https://wttr.in/London?format=3"))
+        .request(&ctx, &HttpRequest::get("https://wttr.in/London"))
```
brooksmtownsend, line 46: Does this break the test?

brooksmtownsend (changes requested):
code looks good, but the `httpclient` test is failing

httpclient/tests/httpclient_test.rs
stevelr: I made that change because it started failing with the format arg on my machine but succeeded without the format arg Worked on my machine, but maybe there's some other problem with the site.  I replaced wttr.in with https://1.1.1.1 which should be faster and more reliable.

stevelr:
Changing the url made the tests pass. I'll merge.

PR #112, opened 2021-09-30
initial checkin sqldb-provider

stevelr:


sqldb/Cargo.toml
```diff
@@ -0,0 +1,43 @@
+[package]
+name = "sqldb"
```
autodidaddict, line 2: By convention, the providers are named for their specific implementation. This crate should be something like `wasmcloud-provider-postgres` - the crate name/artifact name should carry with it the implementation scope.

sqldb/Cargo.toml
```diff
@@ -0,0 +1,43 @@
...
+thiserror = "1.0"
+tokio = { version = "1", features = ["full"] }
+tokio-postgres = { version = "0.7", features=["with-bit-vec-0_6", "with-chrono-0_4","with-uuid-0_8", "with-eui48-0_4"]}
+bb8-postgres = "0.7"
+bit-vec = "0.6"
+uuid = "0.8"
+#postgres-native-tls = "0.5"
+native-tls = "0.2"
+toml = "0.5"
+wasmcloud-interface-sqldb = { path="../../interfaces/sqldb/rust" }
```
autodidaddict, line 31: Assuming this will be swapped with a crates reference when available.

sqldb/src/config.rs
```diff
@@ -0,0 +1,127 @@
...
+        Ok(config)
+    }
+}
+
+pub(crate) async fn create_pool(config: Config) -> Result<crate::Pool, RpcError> {
+    let mut pg_config = tokio_postgres::Config::from_str(&config.uri).map_err(|e| {
+        RpcError::ProviderInit(format!("Invalid db connect string: {}", e.to_string()))
+    })?;
+    pg_config.keepalives_idle(DEFAULT_KEEPALIVE_IDLE);
+    /*
```
autodidaddict, line 90: if we're not intending to support TLS in the initial release, we should delete this comment.

sqldb/src/main.rs
```diff
@@ -0,0 +1,199 @@
...
+}
+
+/// SqlDb - SQL Database connections
+/// To use this capability, the actor must be linked
+/// with "wasmcloud:sqldb"
+/// wasmbus.contractId: wasmcloud:sqldb
+/// wasmbus.providerReceive
+#[async_trait]
+impl SqlDb for SqlDbProvider {
+    async fn execute(&self, ctx: &Context, query: &Query) -> RpcResult<ExecuteResult> {
```
autodidaddict, line 94: See my other comment in the `todo` example review - I think we should add an `execute` wrapper that takes a parameter map which then allows the provider to sanitize each field according to the requirements of the specific database being used, and removes the burden of producing safe/injection-proof SQL statements from the developers.

sqldb/src/main.rs
```diff
@@ -0,0 +1,199 @@
...
+                    query,
+                    &db_err.to_string()
+                );
+                Ok(FetchResult {
+                    error: Some(DbError::from(db_err).into()),
+                    ..Default::default()
+                })
+            }
+        }
+        //Ok(FetchResult::default())
```
autodidaddict, line 190: remove dangling comment

autodidaddict (changes requested):
A few nitpicks, and then suggesting we implement the `execute` wrapper that takes a parameter map and named parameters in the query string so that we (provider) can do safety, escapes, and injection detection instead of the developer.

autodidaddict:
The directory of this provider should be renamed from `sqldb` to `postgres` or `sqldb-postgres`

sqldb/Cargo.toml
stevelr: after it's published to crates.io

sqldb/src/config.rs
stevelr: intend to support TLS in the initial release. I had coded it but it didn't work yet. Instead of debugging, I commented it out to get a stable build and initial handoff for working on petclinic demo. TLS is a must-have feature.

sqldb/src/main.rs
stevelr: prepared statements are also must-have feature. Not sure if they're in scope for first demo release.

sqldb/src/main.rs
stevelr: done

stevelr:
this compiles with latest interface pr

sqldb-postgres/Cargo.toml
```diff
@@ -0,0 +1,44 @@
+[package]
+name = "wasmcloud-provider-sqldb-postgres"
```
autodidaddict, line 2: By convention this should just be wasmcloud-provider-postgres but it's fine this way.

sqldb-postgres/Cargo.toml
```diff
@@ -0,0 +1,44 @@
...
+version = "0.1.0"
+description = "Postgres SQL Database capability provider for contract wasmcloud:sqldb"
+readme = "README.md"
+edition = "2018"
+resolver = "2"
```
autodidaddict, line 7: We should add tags and categories. For tags we usually add at least wasmcloud and I think the category that might apply here is client library (on phone so can't look up the category slug for this atm)

autodidaddict: Ok I looked in https://crates.io/category_slugs and I think the 2 categories we need are wasm and database.

autodidaddict (changes requested):

sqldb-postgres/Cargo.toml
stevelr: we haven't been publishing capability providers to crates.io

autodidaddict: Touche. Move along, nothing to see here.

autodidaddict (approved):

PR #113, opened 2021-10-01
Sqldb cross workflow

stevelr:
- add Cross.toml to sqldb-postgres
- update workflow (for sqldb-postgres) to enable manual workflow trigger (for occasions like this when it's already tagged)


brooksmtownsend (approved):
LGTM

PR #116, opened 2021-11-10
Initial commit - lattice controller provider

autodidaddict:


autodidaddict:
About to head out for the day... @brooksmtownsend @thomastaylor312 I don't know why I got the file not found error while trying to run rust check... maybe I didn't change enough from the provider template GH action yaml?

.github/workflows/lattice-controller.yml
```diff
@@ -0,0 +1,96 @@
...
+    tags:
+    - 'lattice-controller-v*'
+  pull_request:
+    branches: [ main ]
+    paths:
+    - "lattice-controller/**"
+
+env:
+  CARGO_TERM_COLOR: always
+  working-directory: ./provider
```
brooksmtownsend, line 17: @autodidaddict here's the offender for the `file not found`, should be `working-directory: ./lattice-controller`

lattice-controller/Makefile
```diff
@@ -0,0 +1,14 @@
...
+CAPABILITY_ID = wasmcloud:latticecontrol
+VENDOR   = "wasmCloud"
+VERSION  = $(shell cargo metadata --no-deps --format-version 1 | jq -r '.packages[] .version' | head -1)
+REVISION = 0
+oci_url  = localhost:5000/v2/$(PROJECT):$(VERSION)
+
+include ../build/makefiles/provider.mk
+
+test:
+	killall target/debug/kvredis || true
```
brooksmtownsend, line 13: this looks autogenerated but I think it should be `target/debug/latticecontroller` or whatever the resulting binary is

lattice-controller/src/main.rs
```diff
@@ -0,0 +1,406 @@
...
+            (Some(jwt), Some(seed)) => {
+                let kp = KeyPair::from_seed(&seed)
+                    .map_err(|e| RpcError::ProviderInit(format!("key init: {}", e)))?;
+                asynk::Options::with_jwt(
+                    move || Ok(jwt.clone()),
+                    move |nonce| kp.sign(nonce).unwrap(),
+                )
+            }
+            (None, None) => asynk::Options::new(),
+            _ => {
```
brooksmtownsend, line 134: Note: It is possible to do _just_ seed authentication without a JWT and I just added that to `wash` as a part of the context PR (mostly because it enabled my demo of seed auth). Not saying we have to add that quite yet if that isn't an option we want to support, but that is a valid authentication scheme

lattice-controller/src/main.rs
```diff
@@ -0,0 +1,406 @@
...
+    }
+
+    /// Handle notification that a link is dropped
+    async fn delete_link(&self, actor_id: &str) {
+        let mut configs = self.configs.write().await;
+        let _ = configs.remove(actor_id);
+    }
+
+    /// Handle shutdown request
+    async fn shutdown(&self) -> Result<(), Infallible> {
```
brooksmtownsend, line 171: TIL about `Infallible`, love it

brooksmtownsend (changes requested):
Some small requests and notes

stevelr (approved):
Don't forget to add the folder name to the top-level capability-providers/Makefile

lattice-controller/src/main.rs
autodidaddict: right. this is the same set of authentication options the NATS provider supports. we can always upgrade later.

brooksmtownsend (approved):

PR #134, opened 2022-02-02
update dependencies

stevelr:

### DIFF
--- a/nats/Cargo.toml
+++ b/nats/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "wasmcloud-provider-nats"
-version = "0.11.7"
+version = "0.11.8"
 edition = "2021"
 
 [dependencies]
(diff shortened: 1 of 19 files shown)


autodidaddict:
Looks like nearly all the checks are failing, many with failures to find path-based dependencies:

```
Caused by:
  failed to load source for dependency `wasmbus-rpc`
```

nats/Cargo.toml
```diff
@@ -11,7 +11,7 @@ chrono = "0.4"
 crossbeam="0.8"
 futures = "0.3"
 log = "0.4"
-nats = "0.13"
+nats = { version = "0.16.101", package = "nats-aflowt" }
```
brooksmtownsend, line 14: What's the difference between this and just using `nats-aflowt 0.16.103`?

brooksmtownsend (approved):

nats/Cargo.toml
stevelr: it uses nats_aflowt but it's called nats inside the rust code.

stevelr: there's no need to do this anymore, I can back it out. This was before I decided to standardize on aflowt being availble as `wasmbus_rpc::anats`, which also avoids a direct dependency

PR #138, opened 2022-02-14
changed httpclient to use rustls instead of openssl

brooksmtownsend:
This removes the only instance where our capability providers try to dynamically link to libssl and libcrypto
Signed-off-by: Brooks Townsend <brooks@cosmonic.com>

### DIFF
--- a/httpclient/tests/httpclient_test.rs
+++ b/httpclient/tests/httpclient_test.rs
@@ -51,7 +51,8 @@ async fn get_request(_opt: &TestOptions) -> RpcResult<()> {
     assert_eq!(resp.status_code, 200, "status code");
     assert!(resp.header.get("content-type").is_some());
 
-    let body = String::from_utf8_lossy(&resp.body);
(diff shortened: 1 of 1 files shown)


httpclient/tests/httpclient_test.rs
```diff
@@ -46,7 +46,7 @@ async fn get_request(_opt: &TestOptions) -> RpcResult<()> {
     let ctx = Context::default();
 
     let resp = client
-        .request(&ctx, &HttpRequest::get("https://1.1.1.1/"))
+        .request(&ctx, &HttpRequest::get("https://google.com/"))
```
thomastaylor312, line 49: Sigh...that missing IP SAN support

thomastaylor312 (approved):

stevelr:
Good catch on finding ssl deps 

Probably should do case insensitive match:  convert first 1024 bytes to lowercase and then check for `!doctype html`

PR #141, opened 2022-02-24
initial checkin Hashicorp Vault capability provider for KeyValue

stevelr:
Signed-off-by: Steve <steve@cosmonic.com>


kv-vault/README.md
```diff
@@ -0,0 +1,48 @@
...
+| `addr`   | Optional url address for connecting to the vault, such as 'https://server:8200'. The environment variable `VAULT_ADDR` overrides this setting. If neither `addr` nor `VAULT_ADDR` are set, `http://127.0.0.1:8200` is used. |
+| `mount`  | Optional mount point for keyspace. The environment variable `VAULT_MOUNT` overrides this setting. If neither are specified, `secret/` is used.                                                                              | 
+| `certs`  | Optional comma-separated list of files containing CA certificates and/or other TLS client certificates to be loaded. Can also be set with the environment variable `VAULT_CACERT`.                                          |
+
+If either `certs` or `VAULT_CACERT` is set, the provider will use TLS to connect to Vault (and the `addr`(VAULT_ADDR) url should begin with `https:`),
+otherwise TLS will be disabled (and `addr`(VAULT_ADDR) should begin with `http:`).
+
+For convenience, link setting naems may be provided in uppercase or lowercase. Environment variable names are all-caps.
```
autodidaddict, line 20: typo: `naems` -> `names`

kv-vault/README.md
```diff
@@ -0,0 +1,48 @@
...
+| Set             | sets secret to the string value. Internally, uses key as the secret path and stores the string value in a hashmap { date: value }. Can return error if user does not have permission to write to the key path.      |
+| Get             | gets the string value of a secret key. Loads hashmap from key path and returns the data field of the wrapping hashmap. Returns error if the key does not exist or the user does not have access to read the secret. |
+| Contains        | returns true if there is a secret at the key path and it is readable.                                                                                                                                        |
+| Del             | deletes the latest version of the key.                                                                                                                                                                              |
+| SetQuery        | returns the list of secret keys in the requested path.                                                                                                                                                              |
```
autodidaddict, line 35: Kind of a nit.. the purpose of `setquery` on the kv contract is to query the members of a set at a given key. I guess this is sort of applicable, but it feels like a bit of a shoehorn. I'd be fine not supporting set query if this feels like a stretch.

protochron: The goal of that call is to support the equivalent of `vault kv list secret/...`, right? At a first glance `SetQuery` seems like the closest thing we have since you _do_ get back a set (every key would be unique), but it's really a listing operation

autodidaddict (approved):

connorsmith256 (approved):

kv-vault/README.md
stevelr: I agree with the points. The "List" apis in KeyValue would be a worse fit, so if we want to be able to list secrets, this is the closest function. It's sort of a set, defined by the path input parameter.

PR #154, opened 2022-04-28
Blobstore unix file system implementation 

matsbror:
This pull request contains an implementation of the blobstore interface provider for unix file systems. 

Known limitation is that pagination of `ListContainers` or `ListObjects` is not implemented. A reply will return everything.

Asynchronous reply is implemented but not thoroughly tested. 

There are tests for almost all aspects of happy paths of the usage of the interface.

It can also be tested with a test actor: https://github.com/matsbror/fs-test


blobstore-fs/Cargo.toml
```diff
@@ -0,0 +1,26 @@
+[package]
+name = "blobstore-fs"
```
autodidaddict, line 2: There's a naming convention we typically follow when building providers so that they all appear similarly on crates. The name for this package under that convention would be `wasmcloud-provider-blobstore-fs` (the S3 capability provider for this contract is called `wasmcloud-provider-blobstore-s3`.

blobstore-fs/Cargo.toml
```diff
@@ -0,0 +1,26 @@
...
+version = "0.1.0"
+edition = "2021"
+resolver = "2"
+
+[dependencies]
+async-trait = "0.1"
+log = "0.4"
+# wasmcloud-interface-blobstore = "0.2"
+wasmcloud-interface-blobstore = { path = "../../interfaces/blobstore/rust"}
```
autodidaddict, line 11: Obviously prior to publish the changes to the interface need to be published so we can change this back to a crates reference.

blobstore-fs/Makefile
```diff
@@ -0,0 +1,27 @@
+# blobstore-fs Makefile
+
+CAPABILITY_ID = "wasmcloud:blobstore_fs"
```
autodidaddict, line 3: All capability providers implementing the blobstore contract must use the same contract ID, `wasmcloud:blobstore` .

blobstore-fs/Makefile
```diff
@@ -0,0 +1,27 @@
...
+NAME = "Blobstore for unix file system"
+VENDOR = "Wasmcloud"
```
autodidaddict, line 5: convention is `wasmcloud` here.

blobstore-fs/src/main.rs
```diff
@@ -0,0 +1,676 @@
...
+            Ok(s) => Timestamp {
+                sec: s.as_secs() as i64,
+                nsec: 0u32,
+            },
+            Err(e) => return Err(RpcError::InvalidParameter(format!("{:?}", e))),
+        };
+
+        Ok(ContainerMetadata {
+            container_id: arg.clone(),
+            created_at: Some(modified),
```
autodidaddict, line 320: I don't really know why actors need to be aware of the creation time of a particular container, but that's not an issue for this PR, just something to consider for future revisions to the blobstore contract.

blobstore-fs/src/main.rs
```diff
@@ -0,0 +1,676 @@
...
+#[allow(unused)]
+const CAPABILITY_ID: &str = "wasmcloud:blobstore";
+#[allow(unused)]
+const FIRST_SEQ_NBR: u64 = 0;
+
+// main (via provider_main) initializes the threaded tokio executor,
+// listens to lattice rpcs, handles actor links,
+// and returns only when it receives a shutdown message
+//
+fn main() -> Result<(), Box<dyn std::error::Error>> {
```
autodidaddict, line 34: All of our capability providers now create/initialize a tracing subscriber here. It should be an easy set of copy/paste, as you can see an example here https://github.com/wasmCloud/capability-providers/blob/main/blobstore-s3/src/main.rs#L18

autodidaddict (changes requested):
Given that the purpose of a filesystem provider is likely to serve as a test harness on a local workstation where production is running the s3 (or some other) provider, we don't need to block the PR because it doesn't have windows support... but we should definitely add an issue to the repo to indicate that we'll want to have windows support for this eventually... there are a number of tricks we can do with conditional compilation to make this work (the "old" wasmCloud 0.18 runtime has a filesystem provider that worked on windows).

Otherwise it looks good I just highlighted some spots where we want to make things conform to standards.

blobstore-fs/Cargo.toml
matsbror: Will be fixed in next commit.

blobstore-fs/Cargo.toml
matsbror: Yes, absolutely.

blobstore-fs/src/main.rs
matsbror: I did not really ponder on this, but it is a consequence of the interface proposed by @stevelr

blobstore-fs/src/main.rs
matsbror: Done, but what does this do?

matsbror:
The provider has now been tested with Windows to work. I had to change the provider common makefile to make it build cleanly on my machine.

The `make test`, however, does not work in Windows. Not sure if it is intended to.

brooksmtownsend (approved):
Fantastic @matsbror! And thank you for your patience 😄

PR #156, opened 2022-05-11
updated rmp to fix httpserver glibc issue

brooksmtownsend:
Signed-off-by: Brooks Townsend <brooks@cosmonic.com>


stevelr:
since rmp-serde isn't actually used (cargo-udeps says it's the only one that isn't used), I removed it. It passes the build test so far. I'll try merging now to see if it passes cross builds.

PR #160, opened 2022-05-19
Attempt pinning paste instead of rmp for aarch linux

brooksmtownsend:
This PR both pins `paste` to a specific version and removes intermediate `build` and `deps` artifacts between target compilations. It's still unclear as to _why_ this broke suddenly, but it's not currently worth investigating


autodidaddict (approved):
I approve, though I would like to speak with a (Rust dependency ecosystem) manager to lodge a formal complaint.

PR #164, opened 2022-05-31
Run capability provider builds in parallel

brooksmtownsend:
This PR changes the capability provider builds (just httpclient for easier reviews to start) to build in a matrix instead of in sequence. This cuts the release time of a provider to build+test+build provider archive down to 10ish minutes rather than the status quo of 45 minutes.

Here's a previous run: https://github.com/wasmCloud/capability-providers/actions/runs/1919866317
Here's a successful run from this PR before I re-enabled only running release on tag push: https://github.com/wasmCloud/capability-providers/actions/runs/2418640152

### DIFF
--- a/.github/workflows/httpclient.yml
+++ b/.github/workflows/httpclient.yml
@@ -19,19 +19,19 @@ env:
   WASH_SUBJECT_KEY: ${{ secrets.WASMCLOUD_HTTPCLIENT }}
 
 jobs:
-  rust_check:
-    runs-on: ubuntu-latest
-    steps:
-      - uses: actions/checkout@v2
-      - id: run-nats
-        uses: wasmcloud/common-actions/run-nats@main
-      - id: rust-check-action
-        uses: wasmcloud/common-actions/rust-check@main
-        with:
-          working-directory: ${{ env.working-directory }}
+  # rust_check:
+  #   runs-on: ubuntu-latest
+  #   steps:
+  #     - uses: actions/checkout@v2
+  #     - id: run-nats
+  #       uses: wasmcloud/common-actions/run-nats@main
+  #     - id: rust-check-action
+  #       uses: wasmcloud/common-actions/rust-check@main
+  #       with:
+  #         working-directory: ${{ env.working-directory }}
 
   build_artifact:
-    needs: rust_check
+    # needs: rust_check
     # if: startswith(github.ref, 'refs/tags/') # Only run on tag push
     strategy:
       matrix:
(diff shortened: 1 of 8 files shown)


brooksmtownsend:
Marking as ready for review @autodidaddict / @connorsmith256 / @thomastaylor312 just to get feedback on one action, then I can replicate across all actions

.github/workflows/httpclient.yml
```diff
@@ -22,60 +22,131 @@ jobs:
...
+        working-directory: ${{ env.working-directory }}
+
+      - name: Upload executable to GH Actions
+        uses: actions/upload-artifact@v3
+        with:
+          name: ${{ matrix.target }}
+          if-no-files-found: error
+          path: |
+            ${{ env.working-directory }}/target/${{ matrix.target }}/release/${{ env.artifact-name }}
+            ${{ env.working-directory }}/target/${{ matrix.target }}/release/${{ env.artifact-name }}.exe
```
connorsmith256, line 83: Could this be simplified to one line with a wildcard, or are there other files that would match?

```suggestion
            ${{ env.working-directory }}/target/${{ matrix.target }}/release/${{ env.artifact-name }}*
```

.github/workflows/httpclient.yml
```diff
@@ -22,60 +22,131 @@ jobs:
...
+          echo "artifact-name=$(cargo metadata --no-deps --format-version 1 | jq -r '.packages[0].targets[0].name')" >> $GITHUB_ENV
+        working-directory: ${{ env.working-directory }}
+
+      - name: Create provider archive
         working-directory: ${{ env.working-directory }}
-      
+        run: |
+          mkdir -p target/release
+          mv x86_64-unknown-linux-gnu/${{ env.artifact-name }} target/release/
+          make build/${{ env.artifact-name }}.par.gz
```
connorsmith256, line 106: Could this be `make par` for clarity?

connorsmith256 (approved):

.github/workflows/kv-vault.yml
```diff
@@ -92,7 +92,7 @@ jobs:
       - name: Build native executable
         run: |
           chmod +x ~/.cargo/bin/cross
-          cross build --release --target ${{ matrix.target }}
+          XDG_CACHE_HOME=${HOME}/.cache cross build --release --target ${{ matrix.target }}
```
connorsmith256, line 95: Is this directory guaranteed to exist on the runner, or do we need a `mkdir` to be safe?

brooksmtownsend: Good call, I can add a mkdir to be safe

PR #167, opened 2022-06-22
Update/async nats

stevelr:


httpserver-rs/src/lib.rs
```diff
@@ -149,7 +149,12 @@ impl HttpServerCore {
                       query: String| {
                     let inner = Arc::clone(&inner);
                     let linkdefs = linkdefs.clone();
-                    tracing::span::Span::current().record("query", &tracing::field::display(&query));
+                    //let tracer = opentelemetry::global::tracer("HttpServer");
+                    let span = tracing::debug_span!("http request");
+                    span.record("method", &tracing::field::display(method.to_string().as_str()));
+                    span.record("path", &tracing::field::display(path.as_str()));
+                    span.record("query", &tracing::field::display(query.as_str()));
```
connorsmith256, line 156: Hmm, should this handler anonymous function get a name that we can then `instrument`?

connorsmith256 (changes requested):
Providers that need to have their fmt subscriber removed if we're going to enable the otel feature in wasmbus-rpc:
- blobstore-s3
- httpclient
- kv-vault
- lattice-controller
- nats
- sqldb-postgres

I don't see Cargo.lock changes for the last two (nats and sqldb-postgres) either

httpserver-rs/src/lib.rs
stevelr: maybe, but I currently like the level of verbosity on httpserver traces. I think there may be a way to add those to the debug_span! macro to make it simpler. I'll leave this comment unresolved and come back to it later.

stevelr: (fmt subscribers fixed)

connorsmith256: Not sure I follow. Do you mean you like seeing these inline over the `instrument` attribute syntax?

brooksmtownsend (commented):
I've tested and confirmed that lattice-controller, kvredis, nats, httpserver, and httpclient all work. That's sufficient along with the integration tests for me to be confident in the upgrade (once all interfaces + wasmbus-rpc is updated to pinned versions ofc)

lattice-controller/Cargo.toml
```diff
@@ -21,16 +21,16 @@ toml = "0.5"
 tracing = "0.1"
 tracing-subscriber = { version = "0.3", features = ["env-filter"] }
 wascap = "0.8.0"
-wasmbus-rpc = "0.8"
-wasmcloud-control-interface = "0.15.0"
+wasmbus-rpc = { git = "https://github.com/wasmcloud/weld", rev = "6da17f739b0ad0f1a58df2061eedd0d2f92baee0", features = ["otel"] }
+wasmcloud-control-interace = { git = "https://github.com/wasmcloud/control-interface-client", rev ="a9b4570138b6453756735f428277a4af2162b33c" }
```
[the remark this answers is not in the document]
brooksmtownsend: not sure why this duplicated

PR #173, opened 2022-07-18
fix(nats): Adds workpool for handling high-throughput subscriptions

thomastaylor312:
When there is a high volume of messages, the single async task we
spawn does not have the ability to handle the load. This handles
high throughput by spawning the actually message handling onto a
"task pool" gated by semaphores

### DIFF
--- a/nats/Cargo.lock
+++ b/nats/Cargo.lock
@@ -126,9 +126,9 @@ dependencies = [
 
 [[package]]
 name = "async-task"
-version = "4.2.0"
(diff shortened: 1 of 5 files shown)


brooksmtownsend (approved):
For documentation later, can you share any numbers relating to how this improves performance / throughput?

thomastaylor312:
I didn't do any specific benchmarking with this code, but with almost identical code for an internal project, we saw 3-5x speed increases in high throughput situations

PR #175, opened 2022-07-27
chore(*): Cleans up unneeded allocations in our providers

thomastaylor312:
This should allow them to run even slimmer in constrained environments.
Please note I did not bump versions on the providers as there is an
incoming wasmbus-rpc change and there isn't any rush to get these out
right now


stevelr:
at first glance these all look pretty good. Did you run all the tests ('make test' from top-level dir)? blobstore looks like it had the most changes that might affect tests. I agree they shouldn't be version bumped until we roll in wasmbus-rpc 0.10.

blobstore-s3/src/lib.rs
```diff
@@ -644,6 +638,7 @@ impl Blobstore for StorageClient {
                 "cannot put zero-length objects".to_string(),
             ));
         }
+        // TODO: make sure put_object takes an owned `PutObjectRequest` to avoid cloning the whole chunk
```
brooksmtownsend, line 641: Is this something that was intended to make it into this PR?

httpserver-rs/src/settings.rs
```diff
@@ -4,6 +4,7 @@
 //! which determine how the configuration is parsed.
 //!
 //! For the key...
+use std::net::{IpAddr, Ipv4Addr};
```
brooksmtownsend, line 7: should this be with the other imports?

brooksmtownsend (approved):
All looks reasonable 😄

brooksmtownsend:
> at first glance these all look pretty good. Did you run all the tests ('make test' from top-level dir)? blobstore looks like it had the most changes that might affect tests. I agree they shouldn't be version bumped until we roll in wasmbus-rpc 0.10.

@stevelr `make test` is run as a part of the [blobstore s3 action](https://github.com/wasmCloud/capability-providers/runs/7531347044?check_suite_focus=true) now, so it looks like it all passed with flying colors

blobstore-s3/src/lib.rs
thomastaylor312: Yes, it was a note that we need to fix it

httpserver-rs/src/settings.rs
thomastaylor312: Std imports are separated by a line before external crates import, followed by another line before internal imports

PR #178, opened 2022-08-12
Converting provider to multiplex connector

autodidaddict:


lattice-controller/src/client_cache.rs
```diff
@@ -0,0 +1,224 @@
...
+            })
+        }
+        (None, None) => async_nats::ConnectOptions::default(),
+        _ => {
+            return Err(RpcError::InvalidParameter(
+                "must provide both jwt and seed for jwt authentication".into(),
+            ));
+        }
+    };
+    let url = cfg.cluster_uris.get(0).unwrap();
```
connorsmith256, line 164: I'd rather we return an `RpcError` than panic if `cluster_uris` is empty

lattice-controller/src/client_cache.rs
```diff
@@ -0,0 +1,224 @@
...
+    let conn = opts
+        .connect(url)
```
connorsmith256, line 166: Should we add an `event_callback` handler?

https://docs.rs/async-nats/0.18.0/async_nats/struct.ConnectOptions.html#method.event_callback

lattice-controller/src/client_cache.rs
```diff
@@ -0,0 +1,224 @@
...
+) {
+    let expired_keys: Vec<String> = {
+        let meta = m.read().await;
+        let now = Instant::now();
+
+        meta.iter()
+            .filter(|(_k, v)| now.duration_since(v.last_accessed) > period)
+            .map(|(k, _v)| k.to_string())
+            .collect()
+    };
```
connorsmith256, line 188: Thoughts on adding a debug/trace-level message when we evacuate a key?

lattice-controller/src/main.rs
```diff
@@ -51,290 +54,267 @@ impl Default for ConnectionConfig {
...
+                ConnectionConfig {
+                    cluster_uris: vec![arg
+                        .nats_url
+                        .clone()
+                        .unwrap_or_else(|| "0.0.0.0:4222".to_string())],
+                    auth_jwt: arg.user_jwt.clone(),
+                    auth_seed: arg.user_seed.clone(),
+                    lattice_prefix: arg.lattice_id.to_string(),
+                    timeout_ms: 2_000,
+                    auction_timeout_ms: 5_000,
```
connorsmith256, line 123: Should these be configurable?

lattice-controller/src/main.rs
```diff
@@ -51,290 +54,267 @@ impl Default for ConnectionConfig {
...
-        self.lookup_client(ctx)
+        Ok(self
+            .connections
+            .get_client(&arg.lattice_id.to_string())
             .await?
-            .get_host_inventory(&arg.to_string())
+            .get_host_inventory(&arg.host_id)
             .await
-            .map_err(|e| RpcError::Nats(format!("{}", e)))
+            .map(|hi| HostInventory {
```
connorsmith256, line 257: https://tenor.com/NMDa.gif

connorsmith256 (approved):
Some questions/suggestions, but LGTM

lattice-controller/src/client_cache.rs
stevelr: you can use  `let conn = wasmbus_rpc::rpc_client::with_connection_event_logging(opts).connect(url)`
to log on callbacks

lattice-controller/src/main.rs
```diff
@@ -51,290 +54,267 @@ impl Default for ConnectionConfig {
...
+                    },
+                );
+            }
+        }
+
+        client
+            .put_registries(hm)
             .await
-            .map_err(|e| RpcError::Nats(format!("{}", e)))
+            .map_err(|e| RpcError::Nats(format!("{}", e)))?;
```
stevelr, line 158: nit:  instead of `format!("{}",e)` just `e.to_string()`

lattice-controller/src/main.rs
autodidaddict: eventually I think so. not sure if it's required for this version since they were hardcoded previously as well.

lattice-controller/src/client_cache.rs
```diff
@@ -0,0 +1,248 @@
...
+
+    let url = cfg.cluster_uris.get(0).unwrap();
+
+    let conn = opts
+        .event_callback(|event| async move {
+            // lattice prefix/ID will already be on the span from earlier calls
+            match event {
+                async_nats::Event::Disconnect => debug!("NATS client disconnected"),
+                async_nats::Event::Reconnect => debug!("NATS client reconnected"),
+                async_nats::Event::ClientError(err) => debug!("NATS client error occured: {}", err),
```
stevelr, line 179: typo "occured" -> "occurred"

lattice-controller/src/client_cache.rs
```diff
@@ -0,0 +1,248 @@
...
+    m: Arc<RwLock<HashMap<String, ClientMetadata>>>,
+    c: Arc<RwLock<HashMap<String, Client>>>,
+    period: Duration,
+) {
+    let expired_keys: Vec<String> = {
+        let meta = m.read().await;
+        let now = Instant::now();
+
+        meta.iter()
+            .filter(|(_k, v)| now.duration_since(v.last_accessed) > period)
```
stevelr, line 202: this can simplify from `filter(|(_k, v)| now.duration_since(v.last_accessed) > period)`
to `filter(|(_k, v)| v.last_accessed.elapsed() > period)`
.. and you don't need now()

lattice-controller/src/client_cache.rs
```diff
@@ -0,0 +1,248 @@
...
+
+    if !expired_keys.is_empty() {
+        trace!(
+            "Removing NATS clients from cache: {}",
+            expired_keys.join(",")
+        );
+    }
+
+    let mut conns = c.write().await;
+    conns.retain(|k, _v| !expired_keys.contains(k));
```
stevelr, line 215: nit: this does a full table scan and is O(n) for n total items. If len(expired_keys) << len(conns) then it would be more efficient to delete them individually

lattice-controller/src/client_cache.rs
```diff
@@ -0,0 +1,248 @@
...
+        .await
+        .map_err(|e| RpcError::ProviderInit(format!("Nats connection to {}: {}", url, e)))?;
+
+    Ok(conn)
+}
+
+/// Discovers a list of expired (access time within grace period) connections
+/// and then removes them from the cache.
+async fn evacuate_cache(
+    m: Arc<RwLock<HashMap<String, ClientMetadata>>>,
```
stevelr, line 193: I would have preferred dashmap over Hashmap to reduce lock contention, but I see you left a todo to evaluate. 

Please add two more TODOs:
-  evaluate cache crate like https://crates.io/crates/cached which I think would reduce and simplify the code in this crate. 
- increase number of unit tests (not necessary if we switch to `cached` or similar crate if it already has a reasonable test suite).

stevelr (approved):
I would like to see either more unit tests for this, or use of existing and better tested cache crate. Approving now for this sprint but would like to add a backlog task for more unit tests.

lattice-controller/src/client_cache.rs
autodidaddict: it does a full table scan in read mode and then lets go of the lock. this is a deliberate choice to not hold a write lock for any longer than necessary.

lattice-controller/src/client_cache.rs
autodidaddict: I tried dashmap and couldn't get it to work after 3 hours. this is why there's a TODO at the top of this file to explore dashmap further.

lattice-controller/Cargo.toml
```diff
@@ -22,7 +23,9 @@ tracing = "0.1"
 tracing-subscriber = { version = "0.3", features = ["env-filter"] }
 wascap = "0.8.0"
 wasmbus-rpc = { version = "0.9", features = ["otel"] }
-wasmcloud-control-interface = "0.16.2"
+wasmcloud-control-interface = "0.17.0" # lattice client
+wasmcloud-interface-lattice-control = "0.13.1" # provider contract
+#dashmap = "5.3.4"
```
brooksmtownsend, line 28: nit: not used dependency comment

lattice-controller/src/client_cache.rs
```diff
@@ -0,0 +1,248 @@
+use std::{collections::HashMap, sync::Arc};
+
+// TODO: evaluate using dashmap as a replacement for the RwLocks
```
brooksmtownsend, line 3: I revoke my comment if you use dashmap instead :)

lattice-controller/src/main.rs
```diff
@@ -51,290 +54,267 @@ impl Default for ConnectionConfig {
...
-    }
-}
-
 // main (via provider_main) initializes the threaded tokio executor,
 // listens to lattice rpcs, handles actor links,
 // and returns only when it receives a shutdown message
-//
-fn main() -> Result<(), Box<dyn std::error::Error>> {
+#[tokio::main]
+async fn main() -> Result<(), Box<dyn std::error::Error>> {
```
brooksmtownsend, line 61: Because of this change, you must change `provider_start` to `provider_run` to avoid initializing another tokio runtime https://docs.rs/wasmbus-rpc/0.9.3/src/wasmbus_rpc/provider_main.rs.html#123

lattice-controller/src/main.rs
brooksmtownsend: This would be a good "good first issue" to drop in the issue list :)

lattice-controller/README.md
```diff
@@ -1,7 +1,21 @@
 # Lattice Controller Capability Provider
 
-A capability provider that allows actors to interact with the lattice control interface via
-the NATS substrate.
+A capability provider that allows actors to interact with the lattice control interface (`wasmcloud:latticecontrol`) by
+remotely communicating with lattices and the hosts contained within them via NATS.
+
+## Configuration
+This capability provider is designed to facilitate connections to multiple lattices. For each connection that actors need to utilize, there must be an accompanying `set_lattice_credentials` call, _even if you are establishing a connection to the `default` lattice_.
```
brooksmtownsend, line 7: We don't do this in all of our provider READMEs, but since https://crates.io/crates/wasmcloud-interface-lattice-control is no longer compatible with this provider I would love to see just a sample actor function that does something with this provider. This could be at the bottom of the README:
```rust
async fn start_actor(ctx: &Context) -> RpcResult<CtlOperationAck> {
    let lattice = LatticeControllerSender::new();
    // TODO: set lattice creds
    let cmd = StartActorCommand {
        lattice_id: "default".to_string(),
        actor_ref: "wasmcloud.azurecr.io/echo:0.3.4".to_string(),
        annotations: None,
        count: 250,
        host_id: "NB67YNOVU5YB3526RUNCKNZBCQDH2L5NZJKQ6FWOVWGSHNHHEO65RP4A".to_string(),
    };

    debug!(
        "Starting {} instance(s) of actor {} on host {}",
        cmd.count, cmd.actor_ref, cmd.host_id
    );

    lattice.start_actor(ctx, &cmd).await
}
```

brooksmtownsend (changes requested):
Fantastic comments and I really like the rationale behind the choices here. One required request for changing the provider function, a few nits, and one request for an example in the README

lattice-controller/src/client_cache.rs
autodidaddict: I'm going to pass on optimizing this for now because I am pretty sure we're going to ditch this cache in favor of another crate, or at the very least do some other refactoring.

lattice-controller/src/main.rs
autodidaddict: Excellent catch. Thanks.

lattice-controller/README.md
```diff
@@ -1,7 +1,40 @@
...
+This provider no longer supports fallback connections supplied via the provider configuration parameter at startup. In other words, you _must_ invoke `set_lattice_credentials` at least once to use this provider.
+
+
+## Actor Usage Example
+The following is an example of what it looks like for an actor to utilize this capability provider. In this sample, the actor is requesting that the `echo` actor (10 instances of it) be started in the `default` lattice:
+
+```rust
+async fn start_actor(ctx: &Context) -> RpcResult<CtlOperationAck> {
+    let lattice = LatticeControllerSender::new();
+    // NOTE: lattice credentials need to have been set for 'default' before calling this
```
brooksmtownsend, line 27: Can we show the `lattice.set_credentials("default", None, None, None)` call here, noting its for an unauthenticated local lattice?

brooksmtownsend (changes requested):

lattice-controller/README.md
brooksmtownsend: My goal, even if it's pretty self explanatory from the docs, is that someone can come here and take this code as-is to try it

brooksmtownsend (approved):
Looks great!

PR #181, opened 2022-08-17
Added contributing guide and community providers

brooksmtownsend:
This PR seeks to immortalize https://github.com/wasmCloud/wasmCloud/issues/261 with a contributing guide and community providers section in the README.

@matthewtgilbride I'm listing your dynamodb provider as the first example of this, knowing that at this time we've still got a bit of refactoring to do to make it implement the `wasmcloud:keyvalue` contract. Leaving this as a draft, though the Contributing language is ready for review, for now.


CONTRIBUTING.md
```diff
@@ -0,0 +1,16 @@
...
+
+In addition to the above criteria, rigor in code review and standards will be applied when contributing a provider in this way. If you're contributing a first-party capability provider, open a [pull request](https://github.com/wasmCloud/capability-providers/pulls) with your source code as well as adding an entry into the [README.md](https://github.com/wasmCloud/capability-providers/blob/main/README.md).
+
+## Contributing a Community Capability Provider
+In general, community capability providers will be readily accepted as long as they are functional and come with an easy-to-use example. Community capability providers should implement useful functionality, e.g. alternate implementations of HTTPservers or connectors to other services. Demo or one-off providers like [fake payment](https://github.com/wasmCloud/examples/tree/main/provider/fakepay) or factorial are better suited for our examples repository. To submit a community capability provider open a [pull request](https://github.com/wasmCloud/capability-providers/pulls) adding an entry into the [README.md](https://github.com/wasmCloud/capability-providers/blob/main/README.md) under the community section.
```
thomastaylor312, line 16: This paragraph would probably benefit with a definition as to what constitutes a "community capability provider." You provide a nice definition for first party, but not a definition here

CONTRIBUTING.md
```diff
@@ -0,0 +1,16 @@
+# Contributing Guide
+
+The contributing guidelines for this repository follow the same guidelines that are specified for the wasmCloud organization. It's recommended that you read those [rules](https://github.com/wasmCloud/wasmCloud/blob/main/CONTRIBUTING.md) as well as the repository-specific ones below.
+
+As detailed in the [Community Providers RFC](https://github.com/wasmCloud/wasmCloud/issues/261), there are two distinct categories of capability providers that can be contributed: first-party and community.
+## Contributing a wasmCloud First-Party Capability Provider
```
thomastaylor312, line 6: Super nit: Add space above this

README.md
```diff
@@ -28,5 +28,12 @@ The following capability providers are included automatically in every host runt
 
 While neither of these providers requires a _link definition_, to use either of them your actors _must_ be signed with their contract IDs.
 
+## Community Capability Providers
+The following is a list of community supported capability providers developed by members of the wasmCloud community. Please see the [CONTRIBUTING](./CONTRIBUTING.md) guide for information on how to submit your capability provider.
+
+| Provider | Contract | Description |
+| :--- | :--- | :--- |
+| [kvdynamodb](https://github.com/matthewtgilbride/kvdynamodb-provider) | [`wasmcloud:keyvalue`](https://github.com/matthewtgilbride/kvdynamodb-provider) | AWS DynamoDB implementation for the keyvalue contract |
```
matthewtgilbride, line 36: I've just gotten around to giving some TLC to these repos (tests, READMEs).  I think implementing the rest of the `wasmcloud:keyvalue` contract is doable, but what about the _additional_ operation (`keys`) that drove me to create a separate [contract](https://github.com/matthewtgilbride/kvdynamodb)?  

Setting aside the larger questions of how to *version* interfaces and *extend* interfaces, do we think we can reach a consensus on whether `keys` should exist in `wasmcloud:keyvalue` and if so, the details w.r.t. paging?

Happy to take this conversation to a specific issue.

brooksmtownsend: Great question, I think this is something we should bring up in slack or in the community call, because it sounds like that was pretty core to your implementation

brooksmtownsend: We talked about it in the community call today, and the consensus was that the `keys` operation does not belong in `wasmcloud:keyvalue` for two main reasons
1. Some keyvalue stores don't implement key listing functionality (e.g. hashicorp vault, for security purposes)
2. Some keyvalue stores implement key listing functionality in a dangerous way (e.g. Redis single threaded `keys *`)

In order to avoid this ambiguity when selecting a contract, we'll leave it out for now. However, @matthewtgilbride I'd love to get a sense for your use case and figure out if we could tackle the need to list keys in a different way. For example, if we kept track of lists of objects in a keyvalue store, the `keys` requirement may be doable with just get + setadd operations.

We have a call tomorrow where we'll talk about this, so I can update this then as well 😄

jordan-rash:
Would there be a difference in community capability providers that implement wasmcloud contracts as opposed to those that implement new contracts?

thomastaylor312:
@jordan-rash Yes I think we'd have those in community but it would depend on how popular/well-used they are. If someone had a contract like `taylor:catsanddogs` that gave me cute animal pictures, but it just had a single provider with a limited audience, we'd likely say no. Contrast that with a widely-used/popular contract (let's call it `crypto:blockchain` for funsies) across the entire community would probably live a "community contract". If it is popular/common enough I think having the provider implementations "community capability providers" makes sense. Obviously that would be a decision maintainers of this repo could make given all available evidence

Basically we don't want to be a gatekeeper or do anything that even hints at favoritism so most things would be accepted. We just want to have a nicely discoverable source of things people can choose from

jordan-rash:
Thanks @thomastaylor312....I am trying to think about how the best way to do the interfaces would be (from the go module perspective).

While we can install modules via github tags (as opposed to releases), it yields a `go.mod` file that makes it kinda difficult to determine what version of the contract you are utilizing.

While a monorepo would be most convenient, do we think it is the long term solution?  I tend not to think so from my experience with the interfaces in their current state.  There have been a few times when I had to go commit spelunking to figure out what version I was on.

Example 
```
go get github.com/wasmcloud/interfaces/sqldb/tinygo@wasmcloud-interface-sqldb-v0.7.0

// produces

go 1.18

require (
    github.com/wasmcloud/actor-tinygo v0.1.1 // indirect
    github.com/wasmcloud/interfaces/sqldb/tinygo v0.0.0-20220816192629-f6c80bb5212d // indirect
    github.com/wasmcloud/tinygo-cbor v0.1.0 // indirect
    github.com/wasmcloud/tinygo-msgpack v0.1.4 // indirect
)
```

thomastaylor312:
That is a little bit different than @jordan-rash. I thought you were referring to adding them in the readme like this PR does. As for handling this in go. The relative near term is just going to be using `wit` files and codegen. The only Go specific dependency would be the wit-bindgen stuff. Managing those dependencies would likely be done with `wash` for now until we get some standard tooling from the eventual output of the Bytecode alliance SIG Registry. 

Totally willing to talk about this tomorrow in the community meeting

jordan-rash:
Yea my thoughts are a little all over the place, but if this PR is more of a billboard (awesome-wasmcloud_providers) type thing, then I have no further questions your honor.

autodidaddict (approved):

PR #185, opened 2022-09-09
Added shields badges for OCI references

brooksmtownsend:
This is currently requesting a payload from an actor running on Cosmonic 😄 

The source code for the actor is here: https://github.com/brooksmtownsend/ocireffer

Right now this list is hard coded, however, with Azure webhooks and future releases we can update the OCIref automatically


connorsmith256 (approved):
😎

PR #192, opened 2022-10-12
Updated kvredis to support overridable config_json default URL and support retries

brooksmtownsend:
This PR changes the Redis connection to a ConnectionManager which automatically supervises the connection and will reconnect if Redis temporarily disconnects.

If Redis is down, users get an error message like:
```
{"error":"Host send error redis error: Reconnecting failed: Connection refused (os error 61)"}
```

Which is reasonable for the connection being down.

Edit: Smushed another change into this PR which I'm fine breaking out if desired, but I wanted to allow users to override the default URL as config_json. Default functionality is unchanged

### DIFF
--- a/kvredis/src/main.rs
+++ b/kvredis/src/main.rs
@@ -10,7 +10,10 @@
 //!
 use std::{collections::HashMap, convert::Infallible, ops::DerefMut, sync::Arc};
 
-use redis::{aio::Connection, FromRedisValue, RedisError};
+use redis::{
+    aio::{Connection, ConnectionManager},
+    FromRedisValue, RedisError,
+};
 use tokio::sync::RwLock;
 use tracing::{info, instrument, warn};
 use wasmbus_rpc::provider::prelude::*;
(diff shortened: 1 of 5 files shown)


kvredis/src/main.rs
```diff
@@ -73,18 +62,24 @@ impl ProviderHandler for KvRedisProvider {
...
+            // if let Ok(connection) = client.get_async_connection().await {
+            let mut update_map = self.actors.write().await;
+
+            update_map.insert(
+                ld.actor_id.to_string(),
+                RwLock::new(
+                    client
+                        .get_tokio_connection_manager()
+                        .await
+                        .map_err(to_rpc_err)?,
```
connorsmith256, line 70: How friendly are these error messages? Should we prefix them with a friendly string? Also, I thought we didn't want to return RpcErrors in `put_link` anymore

kvredis/src/main.rs
```diff
@@ -39,7 +39,7 @@ fn main() -> Result<(), Box<dyn std::error::Error>> {
 #[services(KeyValue)]
 struct KvRedisProvider {
     // store redis connections per actor
-    actors: Arc<RwLock<HashMap<String, RwLock<Connection>>>>,
+    actors: Arc<RwLock<HashMap<String, RwLock<ConnectionManager>>>>,
```
connorsmith256, line 42: Keeping a map to `Connection`s probably made sense, since I don't think `Connection` is `Clone`. However, `ConnectionManager` is `Clone` which allows for sharing the same underlying TCP connection: https://docs.rs/redis/0.21.6/redis/aio/struct.ConnectionManager.html

Is there a reason we still want a connection manager per actor link, or can we use a single CM?

brooksmtownsend: Great point, I think we definitely want the connectionmanager to accept new connections and we can store the one manager instead

brooksmtownsend: edit: after taking a look this provider actually supports multiple different redis connections so we need a connection manager per actor

kvredis/src/main.rs
brooksmtownsend: great catch, fixed

connorsmith256 (approved):

PR #194, opened 2022-11-10
async-nats updates

stevelr:


sqldb-postgres/src/types.rs
```diff
@@ -64,10 +64,13 @@ where
         Type::UUID => enc.str(&row.get::<'r, usize, uuid::Uuid>(i).to_string()),
 
         // timestamp as iso3339 string in UTC
-        Type::TIMESTAMP | Type::TIMESTAMPTZ => enc.str(
-            &DateTime::<Utc>::from_utc(NaiveDateTime::from_timestamp(row.get(i), 0), Utc)
-                .to_rfc3339(),
-        ),
+        Type::TIMESTAMP | Type::TIMESTAMPTZ => {
```
connorsmith256, line 67: Can this go in its own PR?

connorsmith256 (changes requested):

sqldb-postgres/src/types.rs
stevelr: This change fixes a deprecation warning from the current version of chrono. Anybody who builds this provider (or runs `cargo update` in an existing build) would see the deprecation warning.

stevelr: We've traditionally applied any fixes that result from `cargo update` in the same PR. That practice has pros and cons, and may be a good team discussion topic for another day. I prefer not to change the practice for this particular minor change.

brooksmtownsend (approved):

sqldb-postgres/src/types.rs
thomastaylor312: That context makes a lot more sense. I know from a reviewers perspective, this looked like you were fixing a different bug, but since it was to address an update, I think it is fine

Issue #196, opened 2022-12-05
[BUG] NATS provider does not close connections on `delete_link`


brooksmtownsend:
The code in https://github.com/wasmCloud/capability-providers/blob/main/nats/src/main.rs#L229-L237 makes an assumption that dropping the struct closes the subscription but with testing we've been able to see that the subscription stays alive. This might also result in duplicate subscribers if you re-create the link

vados-cosmonic:
hey @brooksmtownsend I'd like to take a stab at this -- digging into the `nats.rs` docs and I can't find an explicit close/disconnect...

brooksmtownsend:
Fixed with #205

PR #205, opened 2023-02-08
fix: close nats provider connections on link deletion

vados-cosmonic:
This PR addresses the issue (#196) related to NATS client subscriptions not being cleared upon drop of the individual client.

It looks like `nats.rs` does not expose the ability to close all of a clients' subscriptions, so I've added some code to manually track the subscriptions and remove them when a link is being torn down. While not strictly necessary, I've saved the subject names for subscriptions to increase usefulness of error messages.

### DIFF
--- a/nats/src/main.rs
+++ b/nats/src/main.rs
@@ -460,8 +460,8 @@ impl Messaging for NatsMessagingProvider {
 #[cfg(test)]
 mod test {
     use super::*;
-    use wasmbus_rpc::provider::ProviderHandler;
     use crate::ConnectionConfig;
+    use wasmbus_rpc::provider::ProviderHandler;
 
     #[test]
     fn test_default_connection_serialize() {
@@ -496,7 +496,6 @@ mod test {
         assert_eq!(cc3.subscriptions, cc1.subscriptions);
         assert_eq!(cc3.auth_jwt, Some("jawty".to_string()))
     }
-
 
     /// Ensure that unlink triggers subscription removal
     /// https://github.com/wasmCloud/capability-providers/issues/196
--- a/nats/src/main.rs
+++ b/nats/src/main.rs
@@ -153,14 +153,10 @@ struct NatsClientBundle {
     pub sub_handles: Vec<(NatsTopic, JoinHandle<()>)>,
 }
 
-impl NatsClientBundle {
-    fn new(
-        client: async_nats::Client,
-        sub_handles: Vec<(NatsTopic, JoinHandle<()>)>,
-    ) -> NatsClientBundle {
-        NatsClientBundle {
-            client,
-            sub_handles,
+impl Drop for NatsClientBundle {
+    fn drop(&mut self) {
+        for handle in &self.sub_handles {
+            handle.1.abort()
         }
     }
 }
@@ -327,16 +323,8 @@ impl ProviderHandler for NatsMessagingProvider {
             }
         };
 
-        let NatsClientBundle {
-            client,
-            sub_handles,
-        } = self.connect(config, ld).await?;
-
         let mut update_map = self.actors.write().await;
-        update_map.insert(
-            ld.actor_id.to_string(),
-            NatsClientBundle::new(client, sub_handles),
-        );
+        update_map.insert(ld.actor_id.to_string(), self.connect(config, ld).await?);
 
         Ok(true)
     }
@@ -346,21 +334,13 @@ impl ProviderHandler for NatsMessagingProvider {
     async fn delete_link(&self, actor_id: &str) {
         let mut aw = self.actors.write().await;
 
-        if let Some(NatsClientBundle { sub_handles, .. }) = aw.get(actor_id) {
+        if let Some(bundle) = aw.remove(actor_id) {
+            // Note: subscriptions will be closed via Drop on the NatsClientBundle
             debug!(
                 "closing [{}] NATS subscriptions for actor [{}]...",
-                sub_handles.len(),
+                &bundle.sub_handles.len(),
                 actor_id,
             );
-
-            // Stop every subscribe loop related to the actor
-            for (sub, handle) in sub_handles {
-                debug!("ending subscription to subject [{}]...", sub);
-                handle.abort();
-            }
-
-            // Remove the actor from the list
-            aw.remove(actor_id);
         }
 
         debug!("finished processing delete link for actor [{}]", actor_id);
--- a/nats/src/main.rs
+++ b/nats/src/main.rs
@@ -5,7 +5,8 @@ use std::{collections::HashMap, convert::Infallible, sync::Arc, time::Duration};
 use futures::StreamExt;
 use serde::{Deserialize, Serialize};
 use tokio::sync::{OwnedSemaphorePermit, RwLock, Semaphore};
-use tracing::{error, info, instrument, warn};
+use tokio::task::JoinHandle;
+use tracing::{debug, error, instrument, warn};
 use tracing_futures::Instrument;
 use wascap::prelude::KeyPair;
 use wasmbus_rpc::{core::LinkDefinition, otel::OtelHeaderInjector, provider::prelude::*};
@@ -40,19 +41,24 @@ fn main() -> Result<(), Box<dyn std::error::Error>> {
     Ok(())
 }
 
+type NatsTopic = String;
+type ClusterUri = String;
+type AuthJwt = String;
+type AuthSeed = String;
+
 /// Configuration for connecting a nats client.
 /// More options are available if you use the json than variables in the values string map.
 #[derive(Debug, Clone, Serialize, Deserialize)]
 struct ConnectionConfig {
     /// list of topics to subscribe to
     #[serde(default)]
-    subscriptions: Vec<String>,
+    subscriptions: Vec<NatsTopic>,
     #[serde(default)]
-    cluster_uris: Vec<String>,
+    cluster_uris: Vec<ClusterUri>,
     #[serde(default)]
-    auth_jwt: Option<String>,
+    auth_jwt: Option<AuthJwt>,
     #[serde(default)]
-    auth_seed: Option<String>,
+    auth_seed: Option<AuthSeed>,
 
     /// ping interval in seconds
     #[serde(default)]
@@ -136,12 +142,35 @@ impl ConnectionConfig {
     }
 }
 
+/// NatsClientBundles hold a NATS client and information (subscriptions)
+/// related to it.
+///
+/// This struct is necssary because subscriptions are *not* automatically removed on client drop,
+/// meaning that we must keep track of all subscriptions to close once the client is done
+#[derive(Debug)]
+struct NatsClientBundle {
+    pub client: async_nats::Client,
+    pub sub_handles: Vec<(NatsTopic, JoinHandle<()>)>,
+}
+
+impl NatsClientBundle {
+    fn new(
+        client: async_nats::Client,
+        sub_handles: Vec<(NatsTopic, JoinHandle<()>)>,
+    ) -> NatsClientBundle {
+        NatsClientBundle {
+            client,
+            sub_handles,
+        }
+    }
+}
+
 /// Nats implementation for wasmcloud:messaging
 #[derive(Default, Clone, Provider)]
 #[services(Messaging)]
 struct NatsMessagingProvider {
     // store nats connection client per actor
-    actors: Arc<RwLock<HashMap<String, async_nats::Client>>>,
+    actors: Arc<RwLock<HashMap<String, NatsClientBundle>>>,
     default_config: ConnectionConfig,
 }
 
@@ -149,12 +178,12 @@ struct NatsMessagingProvider {
 impl ProviderDispatch for NatsMessagingProvider {}
 
 impl NatsMessagingProvider {
-    /// attempt to connect to nats url (with jwt credentials, if provided)
+    /// Attempt to connect to nats url (with jwt credentials, if provided)
     async fn connect(
         &self,
         cfg: ConnectionConfig,
         ld: &LinkDefinition,
-    ) -> Result<async_nats::Client, RpcError> {
+    ) -> Result<NatsClientBundle, RpcError> {
         let opts = match (cfg.auth_jwt, cfg.auth_seed) {
             (Some(jwt), Some(seed)) => {
                 let key_pair = std::sync::Arc::new(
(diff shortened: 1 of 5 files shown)


vados-cosmonic:
hey @brooksmtownsend would you mind taking a look at this? I'm wondering if you had an inkling it would sort of go this way when the issue was discovered.

It looks like `nats.rs` doesn't actually have a sort of client closing method, so subscription tracking has to be done on our side...

brooksmtownsend:
@vados-cosmonic darn, I could've sworn there was a `drain()` or similar command on the client. I'll double check on the crate and if there's nothing I'll poke the NATS folks in their slack to see what they recommend

nats/src/main.rs
```diff
@@ -168,7 +232,10 @@ impl NatsMessagingProvider {
             // up with 20 subscriptions all getting slammed with up to 75 tasks, but we should wait
             // to do anything until we see what happens with real world usage and benchmarking
             let semaphore = Arc::new(Semaphore::new(75));
-            while let Some(msg) = subscription.next().await {
+
+            let mut shared_subscriber = shared_subscriber.write().await;
+
```
brooksmtownsend, line 237: I think we have to be careful here, the loop below should run for the full length of the provider on behalf of a single subscription, so if we need to get a lock on this subscriber elsewhere.

E.g. I'm concerned this write lock will be held even when we try and clean up the subscription. Just want to verify with testing that's not the case

nats/src/main.rs
```diff
@@ -127,38 +159,70 @@ impl NatsMessagingProvider {
...
         .map_err(|e| {
             error!(subject = %sub, error = %e, "error subscribing subscribing");
             RpcError::Nats(format!("subscription to {}: {}", sub, e))
         })?;
+
+        // Create a shared, rw reference to the subscription
+        // which will be used once at creation and again at destruction
+        let shared_subscriber = Arc::new(RwLock::new(subscriber));
+
+        // Save a cpoy of the sub in the map so it can be unsubscribed when the client is dropped
```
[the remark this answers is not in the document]
vados-cosmonic: oh thanks, will clean that up

nats/src/main.rs
vados-cosmonic: Yeah thanks for noting this -- considering that `delete_link` takes the lock to `actors` (and I didn't introduce any new locks there I assumed that this would be fine... 

What would be the best way to go about testing this? It seems like the tests in `nats_test.rs` would cover this but I can write an explicit test case.

connorsmith256: I think Brooks is right, and as written this code won't allow the provider to close connections when a link is deleted:

- `put_link` calls `connect`, which calls this `subscribe` function
- we spawn a new tokio task which acquires the write lock on `shared_subscriber`. This function runs "forever" (unless the NATS server closes the subscription, or there's an error in the client)
- `delete_link` tries to acquire a write lock on the same subscriber. Since this lock is already being held, we'll never be able to call `unsubscribe`

vados-cosmonic: looking at it, I'm not sure there's a way the listener stops other than by being overwhelmed with messages -- is that right?

I *think* I might have found a way to proivde a spot for the lock to be released:

```rust
            while let Some(msg) = shared_subscriber.write().await.next().await {
                ...
            }
```

Looks like also this is a [[well known dark corner](https://github.com/rust-lang/book/issues/1871) -- I'll switch to `loop`here and explicitly drop, but not quite sure how this will affect perf.

vados-cosmonic: Ah sorry @connorsmith256 didn't see your message there -- yeah looking at it, I  wasn't aware that the loop ran for the entire time/didn't stop -- I've replaced the outer shared sub lock acquisition with a `loop` and explicit `drop` to give other writers a chance

nats/src/main.rs
```diff
@@ -127,38 +178,77 @@ impl NatsMessagingProvider {
...
         .map_err(|e| {
             error!(subject = %sub, error = %e, "error subscribing subscribing");
             RpcError::Nats(format!("subscription to {}: {}", sub, e))
         })?;
+
+        // Create channel for sending control messages to the NATS listen loop for a subscriber
+        let (ctrl_tx, mut ctrl_rx) = mpsc::channel(NATS_LISTEN_LOOP_CTRL_MESSAGE_BUFFER_SIZE);
+
+        // Save a copy of the sub in the map so it can be unsubscribed when the client is dropped
+        let mut _actor_map = self.actors.write().await;
```
thomastaylor312, line 221: Nit: Don't preface with `_` unless you aren't going to use it

nats/src/main.rs
```diff
@@ -91,13 +98,57 @@ impl ConnectionConfig {
...
+    }
+}
+
+/// NATSSubscriberBundles hold a NATS subscriber and
+/// information related to the subscriber
+///
+/// As connecting with a subscriber does not save the subject on which you connected
+struct NATSSubscriberBundle {
+    pub subject: String,
+    pub ctrl_tx: Option<mpsc::Sender<NATSLoopControlMessage>>,
```
thomastaylor312, line 129: If we do go down this route, a [`broadcast` channel](https://docs.rs/tokio/latest/tokio/sync/index.html#broadcast-channel) with a capacity of one is probably the best bet (`oneshot` only allows a single consumer). Then you can clone the receiver multiple times and send the shutdown once to everyone)

nats/src/main.rs
```diff
@@ -168,21 +258,46 @@ impl NatsMessagingProvider {
...
-                    wasmbus_rpc::otel::attach_span_context(&msg);
-                });
-                let permit = match semaphore.clone().acquire_owned().await {
-                    Ok(p) => p,
-                    Err(_) => {
-                        warn!("Work pool has been closed, exiting queue subscribe");
-                        break;
+
+            // Listen for another NATS message to come in
+            loop {
```
thomastaylor312, line 263: So I actually think you might be able to get away with storing the `join_handle` for this spawned task and aborting all of them when the link is deleted. I _think_ it cleans up the underlying future and `Drop`s everything. I did double check and the `Subscription` struct does clean itself up on drop. What I think was causing the bug is that we never stopped this thread. If for some reason it doesn't clean up a single broadcast sender and a cloned receiver is all you'll need (as mentioned in the previous comment). 

Either way, the `actors` data member can be something like `actors: Arc<RwLock<HashMap<String,(Client, Vec<JoinHandle>)>>>,` (or putting the client and list of `JoinHandle`s in a struct). Then you can push a join handle on to that vec.

thomastaylor312:
Left some comments above. Basically I think we can get away with something a bit simpler here. Let me know if you'd like to to clarify anything!

nats/src/main.rs
vados-cosmonic: Right now I'm using a basic `mpsc::channel` -- since there's only one listen loop that's what I need right? one receiver, and multiple senders that could make the rx loop stop.

nats/src/main.rs
vados-cosmonic: I'm not sure what happens when multiple links listen on the same channel, but it looks like we'd get two separate loops on the same channel. I wouldn't want deleting one link to stop all of those right?

Didn't thin of changing `actors` though -- will consider that

nats/src/main.rs
vados-cosmonic: woops yeah will fix this, totally right

nats/src/main.rs
vados-cosmonic: removed in favor of holding `JoinHandle`s

nats/src/main.rs
vados-cosmonic: Removed `_`

nats/src/main.rs
vados-cosmonic: refactored in favor of holding `JoinHandle`s

nats/src/main.rs
vados-cosmonic: This has been refactored to directly use JoinHandles

vados-cosmonic:
Just fixed lint... one more review please, @thomastaylor312 :bow:

nats/src/main.rs
```diff
@@ -32,19 +33,24 @@ fn main() -> Result<(), Box<dyn std::error::Error>> {
     Ok(())
 }
 
+type NatsTopic = String;
```
thomastaylor312, line 36: I'm not a huge fan of aliasing these for purely internal usage, but don't think I want to block on it. @brooksmtownsend any thoughts?

nats/src/main.rs
```diff
@@ -229,11 +286,26 @@ impl ProviderHandler for NatsMessagingProvider {
...
+                "closing [{}] NATS subscriptions for actor [{}]...",
+                sub_handles.len(),
+                actor_id,
+            );
+
+            // Stop every subscribe loop related to the actor
+            for (sub, handle) in sub_handles {
+                debug!("ending subscription to subject [{}]...", sub);
+                handle.abort();
+                drop(handle);
```
thomastaylor312, line 301: No need for the manual drop here. You can remove this. Technically it is dropping at the end of each iteration of the loop

nats/src/main.rs
```diff
@@ -229,11 +286,26 @@ impl ProviderHandler for NatsMessagingProvider {
     #[instrument(level = "info", skip(self))]
     async fn delete_link(&self, actor_id: &str) {
         let mut aw = self.actors.write().await;
-        if aw.remove(actor_id).is_some() {
-            info!("nats closing connection for actor {}", actor_id);
-            // close and drop the connection
-            // dropping the client should close it
-        } // else ignore: it's already been dropped
+
+        if let Some(NatsClientBundle { sub_handles, .. }) = aw.get(actor_id) {
```
thomastaylor312, line 290: Change this to `aw.remove` up here. It will return `Option<T>` anyway and then everything will drop once you've used it. 

Actually, now that I think about it, you can do something like:

```rust
let bundle = {
    let mut aw = self.actors.write().await;
    aw.remove(actor_id)
};
```

Then you can do `if let Some()` here (like you are currently doing) without holding a lock while you clean up

nats/src/main.rs
```diff
@@ -310,3 +398,55 @@ impl Messaging for NatsMessagingProvider {
...
+    ///
+    /// NOTE: this is tested here for easy access to put_link/del_link without
+    /// the fuss of loading/managing individual actors in the lattice
+    #[tokio::test]
+    async fn test_unlink_unsub() {
+        // Build a nats messaging provider
+        let prov = NatsMessagingProvider::default();
+
+        // Actor should have no clients and no subs before hand
+        let actor_map = prov.actors.write().await;
```
thomastaylor312, line 418: Nit (since this is a test), but you should only need a read lock here

nats/src/main.rs
```diff
@@ -310,3 +398,55 @@ impl Messaging for NatsMessagingProvider {
...
+    use super::*;
+    use wasmbus_rpc::provider::ProviderHandler;
+
+    /// Ensure that unlink triggers subscription removal
+    /// https://github.com/wasmCloud/capability-providers/issues/196
+    ///
+    /// NOTE: this is tested here for easy access to put_link/del_link without
+    /// the fuss of loading/managing individual actors in the lattice
+    #[tokio::test]
+    async fn test_unlink_unsub() {
```
thomastaylor312, line 413: This test doesn't actually test that the subs were deleted unfortunately. To check that the sub was deleted, we'd have to talk to nats. The stuff was always getting removed from the hashmap, but we weren't aborting the subscription threads

nats/tests/nats_test.rs
```diff
@@ -141,3 +141,59 @@ async fn send_publish(_opt: &TestOptions) -> RpcResult<()> {
 
     Ok(())
 }
+
+// /// test unlinking during request loop
```
thomastaylor312, line 145: Left over debug?

thomastaylor312 (changes requested):
Just a few more small changes in code and some concerns around tests

nats/src/main.rs
vados-cosmonic: I think it does -- I restructured `delete_link` to force going through the handles and dropping, so the only way you could get delete_link to finish is by clearing out the subs array. which I explicitly check in the test. 

The only way we could get through unsubbing without going through that code path is if something *else* removed the sub before hand (and somehow did so without causing the `drop` of the handle).

This test does not *double check* with NATS, but lexically I think we can be sure that that codepath is run through, given that no other codepath that calls `aw.remove`. There exists no other way for `aw.remove` to be called without going through the codepath that calls `abort()` and the explicit `drop()` (which is superfluous as you mention).

Does that logic make sense?

nats/src/main.rs
vados-cosmonic: I wanted to force a write lock to ensure that no one else could be making changes -- i.e. this test would block until the work in `delete_link` is done.

Does that not make sense?

nats/tests/nats_test.rs
vados-cosmonic: :astonished: Sorry I don't know how I left this in! the test was originally here before our discussion, will remove :bow:

nats/src/main.rs
vados-cosmonic: Sure will do -- and actually dropping *detaches* anyway. Definitely didn't need this.

- [ ] remove explicit `drop` for handle

nats/src/main.rs
vados-cosmonic: See [this comment](https://github.com/wasmCloud/capability-providers/pull/205/files#r1116488085) -- I explicitly made sure to go through the subscriptions lexically because I wanted to make absolutely sure that `.abort()` was called. 

[`drop()`ing a join handle detaches it](https://docs.rs/tokio/latest/src/tokio/runtime/task/join.rs.html#18) as far I can tell, it does not necessarily stop it.

nats/src/main.rs
thomastaylor312: The error as it was occurring before is that nothing was terminating the spawned process. We didn't store join handles and so couldn't abort on drop. So the client was removed from the hashmap, but that didn't actually terminate the spawned tasks. This test does check that we actually remove things when a link is deleted, so that is fine I think, but it doesn't actually do any checking that the subs actually stopped (which we can only do through checking NATS). I think it is good enough for now

nats/src/main.rs
thomastaylor312: Nope, that makes sense. Good idea

nats/src/main.rs
vados-cosmonic: yeah, going through and checking NATS is the :medal_sports: ...

nats/src/main.rs
thomastaylor312: I think I miscommunicated. Your for loop will remain unchanged. But calling read to just get them _and then_ dropping them is superfluous. Remove is also a get, but also pulls it out of the hashmap in one go so you only need to lock once. Then you run through the same `abort` stuff after. If you want complete guarantees, you could wrap the `Vec` in another struct and impl `Drop` for that struct

vados-cosmonic:
Just resolved one round of conflicts... and there's another

nats/src/main.rs
vados-cosmonic: Ahhhh gotcha! thanks will update this

thomastaylor312 (approved):

nats/src/main.rs
brooksmtownsend: Sorry I missed this one, I like the association with types but personally don't use this pattern

PR #210, opened 2023-02-22
Adding support for default connection configuration

autodidaddict:
# Feature or Problem

Unlike many of the other capability providers, this one didn't support the notion of supplying a single default connection configuration for all linked actors, which in turn allows actors to be linked w/out values for convenience.

# Related Issues

None

# Release Information

Next release, currently planning on it being 0.15.1 as this is not a breaking change

# Consumer Impact

If consumers were to accidentally leave connection data blank on link definitions, then instead of crashing this new version of the provider will attempt anonymous localhost connection to the NATS server.

# Testing

Tested on platform(s)
- [ ] x86_64-linux
- [ ] aarch64-linux
- [ ] x86_64-darwin
- [X] aarch64-darwin
- [ ] x86_64-windows

## Platforms

Built on platform(s)
- [ ] x86_64-linux
- [ ] aarch64-linux
- [ ] x86_64-darwin
- [X] aarch64-darwin
- [ ] x86_64-windows

## Unit Test(s)

Added unit test to verify that the connection configuration supplied via the start configuration doesn't need all of the fields. 

## Acceptance or Integration

No automated acceptance tests

## Manual Verification

Added a unit test that verifies that the connection information can be properly deserialized from partial data supplied via JSON in the start provider request. Also manually used `iex` to load this new provider with host configuration and w/out to verify that it could accept the start parameters.

### DIFF
--- a/nats/Cargo.toml
+++ b/nats/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "wasmcloud-provider-nats"
-version = "0.15.1"
+version = "0.16.0"
 edition = "2021"
 
 [dependencies]
--- a/nats/Cargo.toml
+++ b/nats/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "wasmcloud-provider-nats"
-version = "0.15.0"
+version = "0.15.1"
 edition = "2021"
 
 [dependencies]
--- a/nats/src/main.rs
+++ b/nats/src/main.rs
@@ -25,14 +25,7 @@ fn main() -> Result<(), Box<dyn std::error::Error>> {
     // returns when provider receives a shutdown control message
     let host_data = load_host_data()?;
     let provider = if let Some(c) = host_data.config_json.as_ref() {
-        // fall back to the safe default if the JSON is bad
-        let config: ConnectionConfig = match serde_json::from_str(c) {
-            Ok(c) => c,
-            Err(e) => {
-                error!("Failed to parse host data config JSON: {e:?}");
-                ConnectionConfig::default()
-            }
-        };
+        let config: ConnectionConfig = serde_json::from_str(c)?;
         NatsMessagingProvider {
             default_config: config,
             ..Default::default()
--- a/nats/src/main.rs
+++ b/nats/src/main.rs
@@ -25,7 +25,8 @@ fn main() -> Result<(), Box<dyn std::error::Error>> {
     // returns when provider receives a shutdown control message
     let host_data = load_host_data()?;
     let provider = if let Some(c) = host_data.config_json.as_ref() {
-        let config: ConnectionConfig = serde_json::from_str(c)?;
+        let config: ConnectionConfig = serde_json::from_str(c)
+            .expect("JSON deserialization from connection config should have worked");
         NatsMessagingProvider {
             default_config: config,
             ..Default::default()
--- a/nats/src/main.rs
+++ b/nats/src/main.rs
@@ -241,7 +241,13 @@ impl ProviderHandler for NatsMessagingProvider {
         let config = if ld.values.is_empty() {
             self.default_config.clone()
         } else {
-            ConnectionConfig::new_from(&ld.values)?
+            match ConnectionConfig::new_from(&ld.values) {
+                Ok(cc) => cc,
+                Err(e) => {
+                    error!("Failed to build connection configuration: {e:?}");
+                    return Ok(false);
+                }
+            }
         };
         let conn = self.connect(config, ld).await?;
 
--- a/nats/src/main.rs
+++ b/nats/src/main.rs
@@ -58,11 +58,36 @@ struct ConnectionConfig {
     ping_interval_sec: Option<u16>,
 }
 
+impl ConnectionConfig {
+    fn merge(&self, extra: &ConnectionConfig) -> ConnectionConfig {
+        let mut out = self.clone();
+        if !extra.subscriptions.is_empty() {
+            out.subscriptions = extra.subscriptions.clone();
+        }
+        // If the default configuration has a URL in it, and then the link definition
+        // also provides a URL, the assumption is to replace/override rather than combine
+        // the two into a potentially incompatible set of URIs
+        if !extra.cluster_uris.is_empty() {
+            out.cluster_uris = extra.cluster_uris.clone();
+        }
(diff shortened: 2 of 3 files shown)


nats/src/main.rs
```diff
@@ -23,18 +23,32 @@ const ENV_NATS_CLIENT_SEED: &str = "CLIENT_SEED";
...
-        Some("Nats Messaging Provider".to_string()),
-    )?;
+    let host_data = load_host_data()?;
+    let provider = if let Some(c) = host_data.config_json.as_ref() {
+        // fall back to the safe default if the JSON is bad
+        let config: ConnectionConfig = match serde_json::from_str(c) {
+            Ok(c) => c,
+            Err(e) => {
+                error!("Failed to parse host data config JSON: {e:?}");
+                ConnectionConfig::default()
```
connorsmith256, line 33: Should this be a fatal error? I'm not sure a user would want the provider to start with default config if the config they try to pass is malformed

nats/Cargo.toml
```diff
@@ -1,6 +1,6 @@
 [package]
 name = "wasmcloud-provider-nats"
-version = "0.15.0"
+version = "0.15.1"
```
connorsmith256, line 3: Since this is a new feature, I think this should be 0.16.0

connorsmith256 (commented):
This covers the all-or-nothing case. However, if I want to use a common config for the URL, JWT, and seed, but a different topic per link, I'm required to specify common values inside each link. What about renaming to `baseConfig` and using the `values` as overrides?

nats/src/main.rs
autodidaddict: agreed

nats/src/main.rs
```diff
@@ -216,7 +237,12 @@ impl ProviderHandler for NatsMessagingProvider {
     /// If the link is allowed, return true, otherwise return false to deny the link.
     #[instrument(level = "debug", skip(self, ld), fields(actor_id = %ld.actor_id))]
     async fn put_link(&self, ld: &LinkDefinition) -> RpcResult<bool> {
-        let config = ConnectionConfig::new_from(&ld.values)?;
+        // If the link definition values are empty, use the default connection configuration
+        let config = if ld.values.is_empty() {
+            self.default_config.clone()
+        } else {
+            ConnectionConfig::new_from(&ld.values)?
```
connorsmith256, line 244: I think you'll still want to log if this is an error, since the host ignores responses on this topic, right?

autodidaddict: correct as usual

connorsmith256 (changes requested):
> Because that's a feature I didn't need. All I needed was the ability to supply configuration to this provider via the startup configuration. I actually think that might make things more confusing and less predictable. People might not immediately know off hand what the merged configuration is.

Sure, I just thought it'd be a nice improvement. I thought of places I've set up several links between actors and this provider and realized I've pasted the same `NATS_URI` value for all of them

> Also the other providers that take configuration this way, as precedent, don't do config merges they do default-or-linkdef

Which examples are you thinking of? The only one I could find that references config JSON from the host data is redis, which [does "merge"/fall back to the config JSON if the link def doesn't contain a value](https://github.com/wasmCloud/capability-providers/blob/42b53f40278a9076fd21cec0d91372c3a6c72804/kvredis/src/main.rs#L85)

EDIT: actually, I realized this provider is _already_ merging between [whole-provider config and link def config](https://github.com/wasmCloud/capability-providers/blob/42b53f40278a9076fd21cec0d91372c3a6c72804/nats/src/main.rs#L88), it's just doing so in a confusing way. With your changes, we'll no longer do that. Are you sure that won't break anything? I.e. today, if `values` is empty, we still set the default cluster URIs

Do you know why this default of `0.0.0.0:4222` isn't working for you? Is it because it's missing the `nats://` scheme, or because 127.0.0.1 and 0.0.0.0 are different in your environment, or something else? I'm now wondering if there's a simpler change we could make

nats/src/main.rs
```diff
@@ -23,18 +23,25 @@ const ENV_NATS_CLIENT_SEED: &str = "CLIENT_SEED";
 fn main() -> Result<(), Box<dyn std::error::Error>> {
     // handle lattice control messages and forward rpc to the provider dispatch
     // returns when provider receives a shutdown control message
-    provider_main(
-        NatsMessagingProvider::default(),
-        Some("Nats Messaging Provider".to_string()),
-    )?;
+    let host_data = load_host_data()?;
+    let provider = if let Some(c) = host_data.config_json.as_ref() {
+        let config: ConnectionConfig = serde_json::from_str(c)?;
```
connorsmith256, line 28: I missed this before, but since an error here will halt the provider, let's add an `eprintln!` indicating what went wrong (note we can't log here, since `provider_main` hasn't been called yet 😔

nats/src/main.rs
```diff
@@ -216,7 +262,19 @@ impl ProviderHandler for NatsMessagingProvider {
...
     #[instrument(level = "debug", skip(self, ld), fields(actor_id = %ld.actor_id))]
     async fn put_link(&self, ld: &LinkDefinition) -> RpcResult<bool> {
-        let config = ConnectionConfig::new_from(&ld.values)?;
+        // If the link definition values are empty, use the default connection configuration
+        let config = if ld.values.is_empty() {
+            self.default_config.clone()
+        } else {
+            // create a config from the supplied values and merge that with the existing default
+            match ConnectionConfig::new_from(&ld.values) {
+                Ok(cc) => self.default_config.merge(&cc),
```
connorsmith256, line 271: Do we need both `new_from` and `merge`? They're both only referenced once -- here, in the same spot -- so I think the logic from one could move into the other?

connorsmith256 (changes requested):

nats/src/main.rs
autodidaddict: since we're using `?` here, what about an `expect`, which will dump to stderr during said panic?

connorsmith256: Works for me!

nats/src/main.rs
autodidaddict: I looked at this a couple of times and I don't think merging them (see what I did there?) will add that much , other than making the "merge_from" function longer than a page.

connorsmith256 (approved):

PR #211, opened 2023-02-24
Adding sanitize prefilter for host config json

autodidaddict:
# Feature or Problem

Current `main` version of this provider can fail to deserialize JSON with visible escapes, e.g. `\"` appears literally in the string without escaping to a quote. 

# Related Issues

n/a

# Release Information

n/a

# Consumer Impact

Fixes an edge case where some tooling might deliver escape-embedded JSON

# Testing

Tested on platform(s)
- [ ] x86_64-linux
- [ ] aarch64-linux
- [X] x86_64-darwin
- [ ] aarch64-darwin
- [ ] x86_64-windows

## Platforms

Built on platform(s)
- [ ] x86_64-linux
- [ ] aarch64-linux
- [ ] x86_64-darwin
- [X] aarch64-darwin
- [ ] x86_64-windows

## Unit Test(s)

Added a unit test full of pesky newlines and backslash quotes.

## Manual Verification

none (unit test full coverage of change)

### DIFF
--- a/nats/src/main.rs
+++ b/nats/src/main.rs
@@ -25,8 +25,11 @@ fn main() -> Result<(), Box<dyn std::error::Error>> {
     // returns when provider receives a shutdown control message
     let host_data = load_host_data()?;
     let provider = if let Some(c) = host_data.config_json.as_ref() {
-        let config: ConnectionConfig = serde_json::from_str(c)
-            .expect("JSON deserialization from connection config should have worked");
+        // sanitize in case we receive escaped JSON with literal quotes and backslashes
+        let c = unescape::unescape(c).unwrap(); // we want to panic here if we can't do this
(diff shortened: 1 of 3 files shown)


nats/src/main.rs
```diff
@@ -25,8 +25,11 @@ fn main() -> Result<(), Box<dyn std::error::Error>> {
     // returns when provider receives a shutdown control message
     let host_data = load_host_data()?;
     let provider = if let Some(c) = host_data.config_json.as_ref() {
-        let config: ConnectionConfig = serde_json::from_str(c)
-            .expect("JSON deserialization from connection config should have worked");
+        // sanitize in case we receive escaped JSON with literal quotes and backslashes
+        let c = unescape::unescape(c).unwrap(); // we want to panic here if we can't do this
```
thomastaylor312, line 29: Shouldn't we just return the error instead? That should exit the process rather than full panicking

PR #215, opened 2023-02-27
Blocks use of headers on SYS topics

autodidaddict:
# Feature or Problem

When the topic being used is `$SYS.REQ.CLAIMS.UPDATE` (and possibly a few others), the NATS server (currently 2.9.3) will fail to parse the claims update because it's attempting to parse the header instead.

# Related Issues

n/a

# Release Information

nats-v0.16.2

# Consumer Impact

Super small edge case that will only pop up if people are using this provider to perform NATS server/system maintenance

# Testing

re-ran test from Makefile

Tested on platform(s)
- [ ] x86_64-linux
- [ ] aarch64-linux
- [X] x86_64-darwin
- [ ] aarch64-darwin
- [ ] x86_64-windows

Built on platform(s)
- [ ] x86_64-linux
- [ ] aarch64-linux
- [X] x86_64-darwin
- [ ] aarch64-darwin
- [ ] x86_64-windows

## Unit Test(s)

No new unit tests added

## Manual Verification

Pending

### DIFF
--- a/nats/Cargo.lock
+++ b/nats/Cargo.lock
@@ -2875,7 +2875,7 @@ dependencies = [
 
 [[package]]
 name = "wasmcloud-provider-nats"
-version = "0.16.1"
+version = "0.16.2"
 dependencies = [
  "anyhow",
  "async-nats",
--- a/nats/Cargo.toml
+++ b/nats/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "wasmcloud-provider-nats"
-version = "0.16.1"
+version = "0.16.2"
 edition = "2021"
 
 [dependencies]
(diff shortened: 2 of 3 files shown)


brooksmtownsend (approved):
Looks solid, any other edge cases we know about where we should strip headers?

autodidaddict:
Not that I can think of. Ideally we shouldn't have to strip at all, and we don't actually lose anything meaningful by dropping headers on $SYS requests because we know there's no other downstream otel consumers of that message

autodidaddict:
CI test failed, cannot reproduce locally

thomastaylor312 (approved):

PR #218, opened 2023-03-01
build: introduce a workspace

rvolosatovs:
## Feature or Problem

Introduce a workspace to simplify maintenance, speed up builds by
facilitating artifact reuse and minimize dependency graph (and make it
consistent)

Note, a custom resolver introduced in 6c68235 does not seem to be
necessary, since the crate builds just fine without it and is inconsistent with other providers.

Removed to make things simpler with the workspace

## Related Issues

https://github.com/wasmCloud/wasmcloud-test/pull/22

## Testing

Built on platform(s)
- [x] x86_64-linux
- [ ] aarch64-linux
- [ ] x86_64-darwin
- [ ] aarch64-darwin
- [ ] x86_64-windows

Tested on platform(s)
- [x] x86_64-linux
- [ ] aarch64-linux
- [ ] x86_64-darwin
- [ ] aarch64-darwin
- [ ] x86_64-windows

### Manual Verification

Built locally and tested in CI


autodidaddict:
My only concern is the symbolic links in the target folders.  I would rather fix the makefiles and other things than use symlinks and potentially cause problems with portability/windows.

rvolosatovs:
> My only concern is the symbolic links in the target folders. I would rather fix the makefiles and other things than use symlinks and potentially cause problems with portability/windows.

I agree with the concern, however note that the symlinks are removed in #219. If https://github.com/wasmCloud/wasmcloud-test/pull/22 were to be merged and a new version of the crate was released, I could cherry-pick that commit here and there'd be no need for symlinks in `main`

autodidaddict:
okay. I was just looking at this PR in isolation so I didn't know that the symlinks were being removed.

rvolosatovs:
> okay. I was just looking at this PR in isolation so I didn't know that the symlinks were being removed.

see also commit message https://github.com/wasmCloud/capability-providers/pull/218/commits/859c146e2248f1b353f86d6b15ec195819a1a8de

autodidaddict:
> > okay. I was just looking at this PR in isolation so I didn't know that the symlinks were being removed.
> 
> see also commit message [859c146](https://github.com/wasmCloud/capability-providers/commit/859c146e2248f1b353f86d6b15ec195819a1a8de)

Thanks for pointing that out. When I use the review button in Github, I don't see commit mesages.

brooksmtownsend (changes requested):
Looks great, I was concerned this would force us to compile the entire repo for a single provider but that doesn't seem to be the case. 

@rvolosatovs Just to be sure, have you attempted to run `make par-full` in these new directories? I think you will need to add the workspace as a volume in order for the build to resolve the dependency versions. This is a check we run on-release so it might not be caught with our build CI

rvolosatovs:
> @rvolosatovs if you wouldn't mind rebasing and checking out my last comment, I think we can merge this in. Just need to get that addressed

Since we're moving these to `wasmcloud/wasmcloud`, I prefer not to spend more time on this, especially since I'm not familiar with the build process here. Feel free to address any issues

PR #232, opened 2023-05-31
fix(nats-provider): use NATS cluster URI provided by link def

### LINKED ISSUE #231: [BUG] Cannot put URI with NATS provider
This is a bug in the provider's default initialization logic for URIs. The `DEFAULT_URI` is not overridden by the passed in URI.

NATS provider version 0.17.0

Log output: `provider initialization or put-link: NATS connection to 0.0.0.0:4222 `

vados-cosmonic:
## Feature or Problem

Fix the bug where the NATS cluster URI in the link def was being overridden by the default. 

Turns out the arrays were being merged in the `new_from`constructor (`merge` is a bit of a red herring), but only the *first* value is ever used. 

## Related Issues

Resolves #231 

## Release Information

`next`

## Consumer Impact

Customers will be able to specify `URI` in NATS provider linkdefs and they'll take effect.

## Testing

Built on platform(s)
- [x] x86_64-linux
- [ ] aarch64-linux
- [ ] x86_64-darwin
- [ ] aarch64-darwin
- [ ] x86_64-windows

Tested on platform(s)
- [x] x86_64-linux
- [ ] aarch64-linux
- [ ] x86_64-darwin
- [ ] aarch64-darwin
- [ ] x86_64-windows

### Acceptance or Integration

Added a integration test that uses `put_link` with values that *would normally* cause it to fail (to ensure that the values are being used).

### DIFF
--- a/nats/src/main.rs
+++ b/nats/src/main.rs
@@ -124,13 +124,14 @@ impl ConnectionConfig {
         } else {
             ConnectionConfig::default()
         };
+
         if let Some(sub) = values.get(ENV_NATS_SUBSCRIPTION) {
             config
                 .subscriptions
                 .extend(sub.split(',').map(|s| s.to_string()));
         }
         if let Some(url) = values.get(ENV_NATS_URI) {
-            config.cluster_uris.push(url.clone());
+            config.cluster_uris = url.split(',').map(String::from).collect();
         }
         if let Some(jwt) = values.get(ENV_NATS_CLIENT_JWT) {
             config.auth_jwt = Some(jwt.clone());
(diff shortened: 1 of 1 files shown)


nats/src/main.rs
```diff
@@ -206,6 +207,8 @@ impl NatsMessagingProvider {
                 ));
             }
         };
+
+        // Use the first visible cluster_uri
         let url = cfg.cluster_uris.get(0).unwrap();
```
connorsmith256, line 212: Interesting... I'm not sure why we have a vec if we only ever use the first entry 🤔

connorsmith256 (approved):
Awesome, thank you

Issue #238, opened 2023-06-06
Add optional cache-control headers to httpserver responses


thomastaylor312:
This is the same as https://github.com/wasmCloud/examples/issues/216, but in the provider where it should belong. This should be set on the linkdef and used to inject the header if it is set (and the response from the actor doesn't already have the header set)

brooksmtownsend:
Closed with #240

PR #240, opened 2023-06-08
Added optional cache-control to http headers

aish-where-ya:
## Feature or Problem
The user has an option to pass cache-control into the http header as a linked value.

## Related Issues
#238 

## Release Information
Next

## Consumer Impact
No drastic effect. Additional and optional parameter in linkdef value.

## Testing
Manual testing

Built on platform(s)
- [ ] x86_64-linux
- [ ] aarch64-linux
- [ ] x86_64-darwin
- [x] aarch64-darwin
- [ ] x86_64-windows

Tested on platform(s)
- [ ] x86_64-linux
- [ ] aarch64-linux
- [ ] x86_64-darwin
- [x] aarch64-darwin
- [ ] x86_64-windows

### Unit Test(s)
NA

### Acceptance or Integration
NA

### Manual Verification
Manually verified

### DIFF
--- a/httpserver-rs/settings.md
+++ b/httpserver-rs/settings.md
@@ -27,7 +27,7 @@ An empty tls section, or no tls section, disables tls. To enable TLS, both `cert
 - `allowed_headers` - a list of allowed headers, case-insensitive. See [`Access-Control-Allow-Headers`](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Access-Control-Allow-Headers) Default:
 
   ```json
-  ["accept", "accept-language", "content-type", "content-language"]
+  ["accept", "accept-language", "content-type", "content-language","cache-control"]
   ```
 
 - `exposed_headers` - see [`Access-Control-Expose-Headers`](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Access-Control-Expose-Headers)
@@ -63,7 +63,7 @@ Example with all settings
   "cors": {
     "allowed_origins": [],
     "allowed_headers": [
-      "accept", "accept-language", "content-language", "content-type", "x-custome-header"
+      "accept", "accept-language", "content-language", "content-type", "cache-control","x-custome-header"
     ],
     "allowed_methods": [ "GET", "POST", "PUT", "DELETE", "HEAD", "OPTIONS" ],
     "exposed_headers": [],
--- a/httpserver-rs/settings.md
+++ b/httpserver-rs/settings.md
@@ -27,7 +27,7 @@ An empty tls section, or no tls section, disables tls. To enable TLS, both `cert
 - `allowed_headers` - a list of allowed headers, case-insensitive. See [`Access-Control-Allow-Headers`](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Access-Control-Allow-Headers) Default:
 
   ```json
-  ["accept", "accept-language", "content-type", "content-language","cache-control"]
+  ["accept", "accept-language", "content-type", "content-language"]
   ```
 
 - `exposed_headers` - see [`Access-Control-Expose-Headers`](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Access-Control-Expose-Headers)
@@ -42,6 +42,10 @@ For example, the following setting limits uploads to 20 MB.
 ```json
 { "max_content_len": "20M" }
 ```
+
+### Cache Control
+
+An optional set of cache-control values that will appear in the header if they are not already set.
 
 ## Examples of settings files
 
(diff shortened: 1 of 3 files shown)


httpserver-rs/src/lib.rs
```diff
@@ -340,6 +351,7 @@ fn convert_response_headers(
                 continue;
             }
         };
+        map.remove(&name);
```
connorsmith256, line 354: Why was this needed?

httpserver-rs/src/settings.rs
```diff
@@ -186,6 +191,13 @@ impl ServiceSettings {
                 }
             }
         }
+        if let Some(cache_control) = self.cache_control.as_ref() {
+            for val in cache_control.iter() {
+                if http::HeaderValue::from_str(val).is_err() {
+                    errors.push("Invalid Cache Control header".to_string());
```
connorsmith256, line 197: Should we include which one failed/the specific error message?

httpserver-rs/src/settings.rs
```diff
@@ -33,6 +33,7 @@ const CORS_ALLOWED_HEADERS: &[&str] = &[
     "accept-language",
     "content-type",
     "content-language",
+    "cache-control",
```
connorsmith256, line 36: Are these case-sensitive? I noticed we used `Cache-Control` on the response

httpserver-rs/src/settings.rs
```diff
@@ -252,6 +264,16 @@ pub fn load_settings(values: &HashMap<String, String>) -> Result<ServiceSettings
         ));
     }
 
+    // accept cache-control header values
+    if let Some(cache_control) = values.get("cache_control") {
```
connorsmith256, line 268: Let's update the settings documentation as well: https://github.com/wasmCloud/capability-providers/blob/main/httpserver-rs/settings.md

httpserver-rs/src/lib.rs
```diff
@@ -230,7 +229,19 @@ impl HttpServerCore {
                                 http::StatusCode::INTERNAL_SERVER_ERROR
                             }
                         };
-                        *http_response.status_mut() = status;
+                        let http_builder = http::Response::builder()
+                        .status(status);
+                        let http_builder = if let Some(cache_control_header) = arc_inner.settings.cache_control.as_ref(){
+                            let mut builder = http_builder;
+                            for val in cache_control_header{
+                                builder = builder.header("Cache-Control",val)
```
connorsmith256, line 237: Just want to double-check that this appends vs overwrites if there are multiple

connorsmith256 (changes requested):
Just a couple requests for updating docs/better error messages, and then a few questions for my own understanding.

Did we need to split the value and iterate because the http header struct didn't like things like `max-age=604800,must-revalidate`?

Also, how does the `=` interact with wash/the washboard? I.e. `wash ctl link put MXXX VXXX wasmcloud:httpserver cache_control=max-age=604800`

httpserver-rs/src/lib.rs
aish-where-ya: So that if the value is set by the user in the response, that value takes precedence over the one in the linkef. 
@thomastaylor312 can you take a look as well.

httpserver-rs/src/settings.rs
aish-where-ya: Headers are case insensitive

httpserver-rs/src/lib.rs
connorsmith256: How does that translate to the response over HTTP? Can a header appear multiple times, or do these need to be concatenated?

httpserver-rs/settings.md
```diff
@@ -27,7 +27,7 @@ An empty tls section, or no tls section, disables tls. To enable TLS, both `cert
 - `allowed_headers` - a list of allowed headers, case-insensitive. See [`Access-Control-Allow-Headers`](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Access-Control-Allow-Headers) Default:
```
connorsmith256, line 27: GH isn't letting me comment on the file, but I think we need a top-level entry for `cache_control` now, right? These aren't intended to be set as part of the `cors` link def value, or are they?

httpserver-rs/src/lib.rs
aish-where-ya: From what I know, these need to be comma separated instead of individual values. @thomastaylor312 @lachieh may know better.

lachieh: yah, comma separated joined in single header:

`Cache-Control: max-age=0, private, must-revalidate`

connorsmith256 (approved):

Issue #237, opened 2023-06-06
httpserver should have a "readonly" mode that can be used when it is only being used to serve up files or responses


thomastaylor312:
This is the same thing as https://github.com/wasmCloud/examples/issues/217, but in the provider where it belongs. This should probably be a value on the linkdef called `FILESERVER_MODE` with a boolean value

connorsmith256:
$0.02: I would prefer `GETONLY_MODE` or `READONLY_MODE` over specific references to a "fileserver"

thomastaylor312:
I'm good with `READONLY_MODE`

brooksmtownsend:
Closed with https://github.com/wasmCloud/capability-providers/pull/243

PR #243, opened 2023-06-15
Added readonly mode to httpserver

aish-where-ya:
## Feature or Problem
The http server capability provider accepts a value on the linkef named "READONLY_MODE" that accepts boolean values. If this flag is set to true, only GET and HEAD methods are permitted.

## Related Issues
#237 

## Release Information
Next

## Consumer Impact
Additional feature. No significant impact on existing users.

## Testing
Manual testing

Built on platform(s)
- [ ] x86_64-linux
- [ ] aarch64-linux
- [ ] x86_64-darwin
- [x] aarch64-darwin
- [ ] x86_64-windows

Tested on platform(s)
- [ ] x86_64-linux
- [ ] aarch64-linux
- [ ] x86_64-darwin
- [x] aarch64-darwin
- [ ] x86_64-windows

### Unit Test(s)
NA

### Acceptance or Integration
NA

### Manual Verification
Manually verified.

### DIFF
--- a/httpserver-rs/settings.md
+++ b/httpserver-rs/settings.md
@@ -78,8 +78,8 @@ Example with all settings
     "max_age_secs": 300
   },
   "max_content_len": "100M",
-  "cache-control": "max-age=20",
-  "readonly-mode": false,
+  "cache_control": "max-age=20",
+  "readonly_mode": false,
 }
 ```
 
@@ -96,5 +96,3 @@ The link definition connecting an actor to a capability provider includes a key-
   (The option `-w0` omits line breaks within the base64)
 
 - use key `config_json`, with a value that is the raw json. Don't forget to escape quotes.
-
-- use key `readonly_mode`, with a boolean value to set the mode as true or false.
--- a/httpserver-rs/settings.md
+++ b/httpserver-rs/settings.md
@@ -47,6 +47,10 @@ For example, the following setting limits uploads to 20 MB.
 
 An optional set of cache-control values that will appear in the header if they are not already set.
 
+### Read Only Mode
+
+If set to true, it allows only GET and HEAD methods on the provider. Default value is false.
+
 ## Examples of settings files
 
 Bind to all IP interfaces and port 3000, with TLS disabled
@@ -74,7 +78,8 @@ Example with all settings
     "max_age_secs": 300
   },
   "max_content_len": "100M",
-  "cache-control":"max-age=20"
+  "cache-control": "max-age=20",
+  "readonly-mode": false,
 }
 ```
 
@@ -91,3 +96,5 @@ The link definition connecting an actor to a capability provider includes a key-
   (The option `-w0` omits line breaks within the base64)
 
 - use key `config_json`, with a value that is the raw json. Don't forget to escape quotes.
+
+- use key `readonly_mode`, with a boolean value to set the mode as true or false.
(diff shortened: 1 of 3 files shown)


httpserver-rs/src/lib.rs
```diff
@@ -191,7 +191,16 @@ impl HttpServerCore {
                     let span = tracing::debug_span!("http request", %method, path = %path.as_str(), %query);
                     let ld = linkdefs.clone();
                     let arc_inner = arc_inner.clone();
-                    async move {
+                    async move{
+                        if let Some(mode) = ld.values.get("READONLY_MODE"){
```
connorsmith256, line 195: Should this be part of `ServiceSettings`?

connorsmith256 (changes requested):
One question around whether to add this to the typed config on links, and please also update the documentation to reflect the addition of this setting

httpserver-rs/src/lib.rs
brooksmtownsend: If possible, it would be nice to retrieve both the upper and lowercase value here. If that's too verbose, then I would recommend the lowercase since the other settings use that as well.

Can we also add this to settings.md?

brooksmtownsend (changes requested):

httpserver-rs/settings.md
```diff
@@ -74,7 +78,8 @@ Example with all settings
     "max_age_secs": 300
   },
   "max_content_len": "100M",
-  "cache-control":"max-age=20"
+  "cache-control": "max-age=20",
+  "readonly-mode": false,
```
connorsmith256, line 82: These should be snake_case, not kebab-case, right?

httpserver-rs/settings.md
```diff
@@ -90,4 +95,6 @@ The link definition connecting an actor to a capability provider includes a key-
   ```
   (The option `-w0` omits line breaks within the base64)
 
-- use key `config_json`, with a value that is the raw json. Don't forget to escape quotes.
\ No newline at end of file
+- use key `config_json`, with a value that is the raw json. Don't forget to escape quotes.
+
+- use key `readonly_mode`, with a boolean value to set the mode as true or false.
```
connorsmith256, line 100: I just realized this section is really confusing. It doesn't mention any of the other values besides `config_file` and `config_b64` which are used for bulk settings... I would either add the rest of the settings to this section (sorry), or leave this out

connorsmith256 (changes requested):

connorsmith256 (approved):
LGTM!

httpserver-rs/src/settings.rs
```diff
@@ -71,6 +71,9 @@ pub struct ServiceSettings {
     /// cache control options
     pub cache_control: Option<String>,
 
+    /// Flag for read only mode
+    pub readonly_mode: Option<bool>,
+
```
brooksmtownsend, line 76: Does this have to `#[serde(default)]` to maintain compatibility or is it None by default?

brooksmtownsend (approved):

httpserver-rs/src/settings.rs
aish-where-ya: I think it would be better to put `#[serde(default)]` in there

connorsmith256: It's `Some(false)` [by default](https://github.com/aish-where-ya/capability-providers/blob/886a2bc1431a180c4f786e5f5255bd3cec5c55cb/httpserver-rs/src/settings.rs#L102). We don't deserialize this struct unless one of `config_file`, `config_b64`, or `config_json` are provided

We could change the type to be `bool` and simply default to false for clarity, but that's a supernit

PR #260, opened 2023-08-08
ci(httpserver): cross compile httpserver

brooksmtownsend:
## Feature or Problem
After our last round of releases we noticed that we were getting linking errors on _some_ capability providers, but not others, with dependency graphs that matched the exact same releases that failed. We assumed this was because of our use of cross, which has caused us problems in the past. This PR is the first of many that _actually_ sets us up to cross compile these releases. 

This structure of cross compiling and the code came from https://github.com/wasmCloud/wadm/pull/82, huge thanks to @stephenandary for getting this done in a way that worked great for providers too.

You can see a successful run of this action, and download the `provider-archive` for yourself here: https://github.com/wasmCloud/capability-providers/actions/runs/5801109768

## Related Issues
HTTPClient PR, which did fail to release https://github.com/wasmCloud/capability-providers/pull/261

## Release Information
HTTPserver v0.19.1 (I would like to release this so we can test that all functionality works)

## Consumer Impact
I'm removing the `armv7` Linux target from the provider archive because we do not have a host runtime that runs on that target yet.

## Testing

Built on platform(s)
- [ ] x86_64-linux
- [ ] aarch64-linux
- [ ] x86_64-darwin
- [x] aarch64-darwin
- [ ] x86_64-windows

Tested on platform(s)
- [ ] x86_64-linux
- [ ] aarch64-linux
- [ ] x86_64-darwin
- [x] aarch64-darwin
- [ ] x86_64-windows

### Manual Verification
I manually ensured that the httpserver here worked on aarch64-darwin. I want to test this with Linux as well once it releases and I don't have to start from file only.


.github/workflows/httpserver.yml
```diff
@@ -30,61 +30,127 @@ jobs:
...
+        uses: dtolnay/rust-toolchain@stable
+        if: matrix.config.arch != 'aarch64'
         with:
-          name: cross
-          path: ~/.cargo/bin
+          toolchain: stable
+          components: clippy, rustfmt
 
-      - name: Determine artifact name
+      - name: Install latest Rust stable toolchain
```
connorsmith256, line 90: Looks like this step is duplicated?

brooksmtownsend: touche

connorsmith256 (approved):

stephenandary:
No worries, and thanks for shout out. That was based on work that @thomastaylor312 had shared that I used to do that, so thanks Taylor lol.

PR #269, opened 2023-09-26
Initial implementation of automatic service discovery

autodidaddict:
This adds a few keys to the link definition that will allow an actor to expose its request/reply services as "NATS micro" discoverable services.

This is currently not something that should be put into production because the current version of async-nats's `service` implementation doesn't match the spec so the 2.10-aware `nats` CLI will fail to deserialize payloads from these services.

### DIFF
--- a/nats/src/config.rs
+++ b/nats/src/config.rs
@@ -1,0 +1,145 @@
+use serde::{Deserialize, Serialize};
+use std::collections::HashMap;
+use wasmbus_rpc::error::{RpcError, RpcResult};
+
+const DEFAULT_NATS_URI: &str = "0.0.0.0:4222";
+const ENV_NATS_SUBSCRIPTION: &str = "SUBSCRIPTION";
+const ENV_NATS_URI: &str = "URI";
+const ENV_NATS_CLIENT_JWT: &str = "CLIENT_JWT";
+const ENV_NATS_CLIENT_SEED: &str = "CLIENT_SEED";
+const ENV_SERVICE_NAME: &str = "SERVICE_NAME";
+const ENV_SERVICE_ENDPOINTS: &str = "SERVICE_ENDPOINTS";
+const ENV_SERVICE_DESCRIPTION: &str = "SERVICE_DESCRIPTION";
+const ENV_SERVICE_VERSION: &str = "SERVICE_VERSION";
+
+/// Configuration for connecting a nats client.
+/// More options are available if you use the json than variables in the values string map.
+#[derive(Debug, Clone, Serialize, Deserialize, PartialEq)]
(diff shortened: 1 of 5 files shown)


autodidaddict:
Note, the patch that I am awaiting before this PR should be merged can be found here: https://github.com/nats-io/nats.rs/commit/02fd08693608023fe460a170ee30bec3c2ffefd8

It should be available in whatever `async-nats` release is > 0.32

nats/src/main.rs
```diff
@@ -322,6 +306,7 @@ impl ProviderHandler for NatsMessagingProvider {
     #[instrument(level = "debug", skip(self, ld), fields(actor_id = %ld.actor_id))]
     async fn put_link(&self, ld: &LinkDefinition) -> RpcResult<bool> {
         // If the link definition values are empty, use the default connection configuration
+        eprintln!("CONFIG: {ld:?}");
```
brooksmtownsend, line 309: Should remove later once we're done with testing 🙂

nats/src/config.rs
```diff
@@ -0,0 +1,145 @@
...
+            return Err(RpcError::InvalidParameter(
+                "if you specify jwt, you must also specify a seed".to_string(),
+            ));
+        }
+
+        if config.cluster_uris.is_empty() {
+            config.cluster_uris.push(DEFAULT_NATS_URI.to_string());
+        }
+
+        eprintln!("{config:?}");
```
brooksmtownsend, line 141: Same comment re: removing once finished testing

nats/src/services.rs
```diff
@@ -0,0 +1,30 @@
...
+}
+
+pub(crate) async fn dispatch_request_waiter(subject: &str, bytes: Bytes) {
+    let mut waiters = REQUEST_WAITERS.write().await;
+    if let Some(sender) = waiters.remove(subject) {
+        if let Err(_) = sender.send(bytes) {
+            warn!("Receiver side of request waiter dropped");
+        }
+    }
+}
```
brooksmtownsend, line 25: Is the request_waiter pattern a new recommended pattern in NATS to respond to incoming requests in a work-limited queue? I think I understand what's going on, it just seems a little complicated to deal with this lazy static to send responses

autodidaddict: This was a bit of an annoying aspect of implementing this. Keeping track of the waiters is basically required to correlate the replies coming from the actor with pending requests so that the NATS rust client can then use the `Respond` function.

If the stats/services API thing did this at a lower level, we wouldn't have to maintain this lookup table here.

autodidaddict: on the other side of the `sender.send` is a waiter getting ready to call `request.Respond`

PR #271, opened 2023-09-27
fix: Blobstore FS and S3 provider fixes

brooksmtownsend:
## Feature or Problem
This PR fixes two possible low severity issues in the blobstore providers. In the FS provider we're preventing infinite filesystem recursion, and in the S3 provider we're preventing an infinite loop when an actor send back invalid responses for `send_chunk`.

## Related Issues
N/A

## Release Information
blobstore-fs 0.5.1
blobstore-s3 0.5.1

## Testing

Built on platform(s)
- [ ] x86_64-linux
- [ ] aarch64-linux
- [ ] x86_64-darwin
- [x] aarch64-darwin
- [ ] x86_64-windows

Tested on platform(s)
- [ ] x86_64-linux
- [ ] aarch64-linux
- [ ] x86_64-darwin
- [ ] aarch64-darwin
- [ ] x86_64-windows

### DIFF
--- a/blobstore-fs/src/fs_utils.rs
+++ b/blobstore-fs/src/fs_utils.rs
@@ -3,7 +3,10 @@ use std::vec::Vec;
 
 /// Traverses a file system starting at location `root` and returning a list of all directories
 /// contained in that directory, recursively, relative to the original root at level 0.
-pub fn all_dirs(root: &Path, prefix: &Path) -> Vec<PathBuf> {
+pub fn all_dirs(root: &Path, prefix: &Path, depth: u32) -> Vec<PathBuf> {
+    if depth > 1000 {
+        return vec![];
+    }
     let mut dirs: Vec<PathBuf> = match std::fs::read_dir(root) {
         Ok(rd) => rd
             .filter(|e| match e {
@@ -23,7 +26,7 @@ pub fn all_dirs(root: &Path, prefix: &Path) -> Vec<PathBuf> {
     // Now recursively go in all directories and collect all sub-directories
     let mut subdirs: Vec<PathBuf> = Vec::new();
     for dir in &dirs {
-        let mut local_subdirs = all_dirs(prefix.join(dir.as_path()).as_path(), prefix);
+        let mut local_subdirs = all_dirs(prefix.join(dir.as_path()).as_path(), prefix, depth + 1);
         subdirs.append(&mut local_subdirs);
     }
     dirs.append(&mut subdirs);
@@ -54,7 +57,7 @@ mod tests {
             );
         }
 
-        let dirs = all_dirs(root, root);
+        let dirs = all_dirs(root, root, 0);
 
         clear_state(root);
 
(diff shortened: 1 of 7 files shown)


blobstore-fs/src/fs_utils.rs
```diff
@@ -3,7 +3,10 @@ use std::vec::Vec;
 
 /// Traverses a file system starting at location `root` and returning a list of all directories
 /// contained in that directory, recursively, relative to the original root at level 0.
-pub fn all_dirs(root: &Path, prefix: &Path) -> Vec<PathBuf> {
+pub fn all_dirs(root: &Path, prefix: &Path, depth: u32) -> Vec<PathBuf> {
+    if depth > 1000 {
+        return vec![];
```
connorsmith256, line 8: Returning an empty list is a little confusing. Should this function return a `Result<Vec<PathBuf>>` instead?

Also, while you're here.... We shouldn't be panicking below, when failing to read a directory

blobstore-s3/src/lib.rs
```diff
@@ -200,7 +201,13 @@ impl StorageClient {
                         object_id: cobj.object_id.clone(),
                     },
                 )
-                .await?;
+                .await?,
+            ) else {
+                return Err(RpcError::InvalidParameter(
```
connorsmith256, line 206: `send_chunk` is already returning an `RpcError::Rpc`. Should we simply return that instead? I don't think this error message is accurate

connorsmith256 (changes requested):

blobstore-fs/src/fs_utils.rs
brooksmtownsend: This is recursive so what it'll do is return an empty list for directories when the depth is over 1000, but we're actually accumulating the list below

blobstore-s3/src/lib.rs
brooksmtownsend: It may or may not return an `RpcError::Rpc` if the sent chunks is 0, right? E.g. operation to the actor is successful but they return a response with 0 bytes

blobstore-fs/src/fs_utils.rs
brooksmtownsend: _realistically_ anyways this is only hit when someone lists a directory with one _thousand_ nested folders beneath it and they have a need for it

connorsmith256: Ah, thank you 👍

blobstore-s3/src/lib.rs
connorsmith256: Hmm, I might have misunderstood. So we're still early returning if `send_chunk` returns an error, and all you're adding here is a different error case to handle when `send_chunk` returns 0?

connorsmith256: Got it, thanks 👍

connorsmith256 (approved):

PR #278, opened 2024-01-27
feat: add custom CA support to the NATS provider, rework Redis TLS

protochron:
## Feature or Problem
This PR does two things:
* adds the ability to add a custom root CA to each NATS linkdef so that you can connect to servers that might be secured with self-signed certs
* reworks how the Redis provider depends on rustls and other dependencies so that the webroot PKI certs are embedded. This removes a dependency on the host for anything TLS related.

### DIFF
--- a/nats/src/main.rs
+++ b/nats/src/main.rs
@@ -24,6 +24,7 @@ const ENV_NATS_SUBSCRIPTION: &str = "SUBSCRIPTION";
 const ENV_NATS_URI: &str = "URI";
 const ENV_NATS_CLIENT_JWT: &str = "CLIENT_JWT";
 const ENV_NATS_CLIENT_SEED: &str = "CLIENT_SEED";
+const ENV_NATS_TLS_CA: &str = "TLS_CA";
 
 fn main() -> Result<(), Box<dyn std::error::Error>> {
     // handle lattice control messages and forward rpc to the provider dispatch
@@ -74,6 +75,8 @@ struct ConnectionConfig {
     auth_jwt: Option<String>,
     #[serde(default)]
     auth_seed: Option<String>,
+    #[serde(default)]
+    tls_ca: Option<String>,
 
     /// ping interval in seconds
     #[serde(default)]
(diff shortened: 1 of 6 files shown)


nats/src/main.rs
```diff
@@ -154,6 +157,9 @@ impl ConnectionConfig {
         if config.cluster_uris.is_empty() {
             config.cluster_uris.push(DEFAULT_NATS_URI.to_string());
         }
+        if let Some(tls_ca) = values.get("tls_ca") {
```
connorsmith256, line 160: Nit: can we give this a name and store it next to [the others](https://github.com/wasmCloud/capability-providers/blob/main/nats/src/main.rs#L26)?

nats/src/main.rs
```diff
@@ -485,6 +473,64 @@ fn should_strip_headers(topic: &str) -> bool {
...
+        }
+        (None, None) => async_nats::ConnectOptions::default(),
+        _ => {
+            return Err(RpcError::InvalidParameter(
+                "must provide both jwt and seed for jwt authentication".into(),
+            ));
+        }
+    };
+
+    if let Some(tls_ca) = &cfg.tls_ca {
```
connorsmith256, line 496: Nit:

```rust
if let Some(tls_ca) = &cfg.tls_ca {
  add_tls_ca(tls_ca, opts)
} else {
  Ok(opts)
}
```

nats/src/main.rs
```diff
@@ -485,6 +473,64 @@ fn should_strip_headers(topic: &str) -> bool {
...
+                store_builder.add_parsable_certificates(&[cert]);
+            }
+            _ => {
+                return Err(RpcError::ProviderInit(
+                    "tls ca: invalid certificate type, must be a DER encoded PEM file".into(),
+                ));
+            }
+        }
+
+        let tls_client = async_nats::rustls::ClientConfig::builder()
```
connorsmith256, line 522: Nit: since the other branches all return, moving this to the end of the function would make it easier to read

connorsmith256 (approved):
Nits only

connorsmith256 (approved):

PR #281, opened 2024-05-23
ci(*): pin to 0.26 for par creation

brooksmtownsend:
Signed-off-by: Brooks Townsend <brooksmtownsend@gmail.com>

## Feature or Problem
This PR pins to wash 0.26 for provider archive generation as these providers are only used with pre-1.0 wasmCloud and need the capability contract ID.


protochron (approved):
LGTM

